PR #35, opened 2026-09-01
feat: make the agent title a configurable format string

yuensunn:
## What

Adds `HERDR_AUTO_TITLE_AGENT_FORMAT` — a format string for how an agent pane's title reads. Default `{agent} › {activity}` reproduces the current behavior exactly.

- `{agent}` — the agent name (e.g. `claude`)
- `{activity}` — the work

Examples:
- `{activity}` — drop the agent prefix (for a terminal whose own tab HUD already names the agent)
- `{activity} ({agent})` — reorder

A format missing `{activity}` is rejected with a warning and the default kept.

## Why

The agent name is not always wanted in the tab title — some terminals/HUDs already show which agent a pane holds, making `claude ›` redundant. A format string lets each user decide, without a new boolean per preference.

## Notes

- Default preserves existing behavior.
- Agent-panes only; non-agent titles are untouched.
- Tests + README/configuration docs included; `go test ./...`, `go vet`, `gofmt` clean.
- Commit history has a few steps (boolean → format string → trim); happy to squash to a single commit on request.

### DIFF
--- a/README.md
+++ b/README.md
@@ -124,6 +124,7 @@ restarts**, with the same `herdr server stop` the install needs.
 | `HERDR_AUTO_TITLE_POSITION`    | `true`                                    | Put each tab's position in front of its title                              |
 | `HERDR_AUTO_TITLE_MANUAL_FILE` | `manual-names.json`, next to `config.env` | Where tabs you renamed by hand are remembered; empty keeps them in memory  |
 | `HERDR_AUTO_TITLE_TRANSCRIPT`  | `true`                                    | Read an agent's own session transcript when it has not titled its terminal |
+| `HERDR_AUTO_TITLE_AGENT_FORMAT` | `{agent} › {activity}`                   | How an agent pane's title reads. `{agent}` is the agent, `{activity}` the work. Drop `{agent}` (say `{activity}`) for a terminal whose own tab HUD already names the agent. Must keep `{activity}` |
 
 ## Documentation
 
--- a/internal/app/config.go
+++ b/internal/app/config.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"path/filepath"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/joho/godotenv"
@@ -18,13 +19,14 @@ import (
 // Environment variables Auto Title reads. The configuration file holds the
 // same names; see docs/architecture/configuration.md.
 const (
-	EnvDebug      = "HERDR_AUTO_TITLE_DEBUG"
-	EnvPoll       = "HERDR_AUTO_TITLE_POLL_MS"
-	EnvMaxLength  = "HERDR_AUTO_TITLE_MAX_LENGTH"
-	EnvBranchMax  = "HERDR_AUTO_TITLE_BRANCH_MAX"
-	EnvPosition   = "HERDR_AUTO_TITLE_POSITION"
-	EnvManual     = "HERDR_AUTO_TITLE_MANUAL_FILE"
-	EnvTranscript = "HERDR_AUTO_TITLE_TRANSCRIPT"
+	EnvDebug       = "HERDR_AUTO_TITLE_DEBUG"
+	EnvPoll        = "HERDR_AUTO_TITLE_POLL_MS"
+	EnvMaxLength   = "HERDR_AUTO_TITLE_MAX_LENGTH"
+	EnvBranchMax   = "HERDR_AUTO_TITLE_BRANCH_MAX"
+	EnvPosition    = "HERDR_AUTO_TITLE_POSITION"
+	EnvManual      = "HERDR_AUTO_TITLE_MANUAL_FILE"
+	EnvTranscript  = "HERDR_AUTO_TITLE_TRANSCRIPT"
+	EnvAgentFormat = "HERDR_AUTO_TITLE_AGENT_FORMAT"
 )
 
 // ConfigFile is the configuration file, read from the same directory the
@@ -56,6 +58,12 @@ type Config struct {
 	// when the agent has not titled its terminal. It reads what the user has
 	// been saying to that agent, so it can be turned off.
 	ReadTranscripts bool
+	// AgentFormat shapes how an agent pane's title joins the agent's name to
+	// its activity. `{agent}` is the agent's name, `{activity}` is the work, so
+	// the default `{agent} › {activity}` reads `claude › <activity>`. Dropping
+	// `{agent}` (say `{activity}`) leaves the activity on its own, which suits
+	// a terminal whose own tab HUD already names the agent.
+	AgentFormat string
 }
 
 // LoadConfig reads configuration from the configuration file and the
@@ -74,6 +82,7 @@ func LoadConfig() (Config, []string) {
 		ShowPosition:    true,
 		ManualPath:      state.DefaultManualPath(),
 		ReadTranscripts: true,
+		AgentFormat:     resolver.DefaultAgentFormat,
 	}
 
 	cfg.Debug = fromEnv(&warnings, EnvDebug, cfg.Debug, boolean)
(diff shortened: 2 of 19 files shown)


kryptamine:
Thanks for the contribution! I really like the idea of using templates here. I'll take a look at it today after work.

internal/resolver/resolver.go
```diff
@@ -114,12 +125,12 @@ func New(maxLength int, sources ...Source) *Deterministic {
...
 // Default builds the resolver Auto Title ships with, so the binary and the
 // tests cannot drift apart on what the chain contains.
-func Default(maxLength, branchMax int) *Deterministic {
+func Default(maxLength, branchMax int, agentFormat string) *Deterministic {
 	return New(maxLength,
-		NewAgent(),
-		NewTerminalTitle(),
-		NewTranscript(),
-		NewProcess(),
+		NewAgent(agentFormat),
```
kryptamine, line 130: **agentFormat is threaded through four sources, none of which decides anything with it**

Agent, TerminalTitle, Transcript and Process each gained the same field and the same constructor parameter. All four do exactly one thing with it: hand it on to activityFrom / formatAgent.

The cost:

- four struct fields, four constructor signatures, an extra parameter on activityFrom;
- 52 Default( call sites in tests rewritten purely to carry a constant through;
- those calls then blew past golines max-len: 100, and the formatter unfolded them into ).Resolve(tabWithPane(pane)).Name

The format is a property of title assembly, not of any source. The package already has the canonical home for exactly this kind of value: Deterministic holds maxLength for the same reason.

The move: give Parts an Agent field and apply the format once, in collect, right where a source hands its result back - the same point in the flow as today, but in one place instead of four:

```go
type Parts struct {
      Context string
      Branch  string
      // Agent names the agent that reported the activity. It is kept apart from
      // the activity because the user's format decides whether it is shown.
      Agent    string
      Activity string
}

for _, source := range d.sources {
      parts, ok := source.Resolve(pane)
      if !ok {
              continue
      }

      // The format joins Agent to Activity; without {agent} and without an
      // activity nothing is left, and the source yields to the next one.
      parts = d.format.apply(parts)
      if parts == (Parts{}) {
              continue
      }

      found.take(source, parts)
      ...
}
```
activityFrom then returns Parts{Agent: kind, Activity: stripKind(activity, kind)} for an agent pane and Parts{Activity: qualify(activity, kind)} otherwise, as it does now. Sources go back to struct{}, take and complete() are untouched, and all 52 test call sites revert.

internal/resolver/process.go
```diff
@@ -100,23 +100,61 @@ func stripKind(detail, kind string) string {
...
+func (p Process) Resolve(pane *state.PaneState) (Parts, bool) {
 	kind := paneKind(pane)
 	if kind == "" {
 		return Parts{}, false
 	}
 
+	// An agent pane named here has no activity to report, so the format decides
+	// whether the agent's name alone still names the tab. A format without
+	// {agent} leaves nothing, and the pane falls through to a lower source.
+	if pane.HasAgent() {
```
kryptamine, line 150: **Process.Resolve uses the formatter as a predicate**

A source that was five lines now calls the formatter with a knowingly empty activity in order to ask a question about the format. This is ad-hoc branching bolted into an unrelated flow: Process is not about agents or templates, it is about which program is running in the pane.
After fixing https://github.com/kryptamine/herdr-auto-title/pull/35/changes#r3906008004 this branch disappears entirely

internal/app/config.go
```diff
@@ -160,6 +171,17 @@ func boolean(raw string) (bool, error) {
 	return value, nil
 }
 
+// agentFormat keeps the title template a user can shape, but only once it still
+// carries the agent's work: a format without {activity} would drop what the
+// title is for, so it is rejected and the default kept.
+func agentFormat(raw string) (string, error) {
+	if !strings.Contains(raw, "{activity}") {
```
kryptamine, line 178: **The template grammar lives in two packages and is re-parsed on every poll**

internal/app/config.go:178 knows {activity} is a placeholder. internal/resolver owns DefaultAgentFormat (resolver.go:24), the substitution, and separately knows about {agent} (process.go:119). Neither package owns the grammar.

Consequences:

- strings.Contains(format, "{agent}") and strings.NewReplacer(...) run per agent pane, per poll (2 Hz by default). Not a disaster, but CLAUDE.md accounts for the poll loop's cost down to fractions of a millisecond, and this is allocation for a constant known at startup.
- A string sniff drives control flow: !strings.Contains(format, "{agent}") decides whether a source yields. The invariant is implicit.

The move - a typed value in the package that owns the grammar:
```go
// AgentFormat is how an agent pane's title joins the agent's name to its work.
type AgentFormat struct {
      replacer   *strings.Replacer
      showsAgent bool
}
```
func ParseAgentFormat(raw string) (AgentFormat, error) // rejects a format without {activity}

Config.AgentFormat becomes resolver.AgentFormat, errNoActivity moves next to the grammar, and an invalid format becomes structurally unable to reach the resolver, rather than relying on config having checked.

internal/resolver/process.go
```diff
@@ -100,23 +100,61 @@ func stripKind(detail, kind string) string {
 	return strings.Trim(trimmed, " -–—|:"+separatorRune)
 }
 
+// formatAgent joins an agent pane's name to its activity through the user's
+// format: {agent} is the name, {activity} is the work. The default
+// `{agent} › {activity}` reproduces what a title has always shown.
+//
+// The substituted values are already sanitized at their source, so this only
+// fills the format; Format sanitizes the assembled title as a whole, which is
```
kryptamine, line 108: **The dangling-separator guarantee does not hold**

▎ Format sanitizes the assembled title as a whole, which is what collapses and trims a separator that a substituted-empty field left dangling

That is only true for the built-in ›: trimmable in sanitize.go:24 is " ›" and nothing else. Running a silent agent (agent present, no activity) through several formats:

{agent} › {activity}   ->  "dashboard › claude"      ok
{agent}: {activity}    ->  "dashboard › claude:"     broken
{agent} - {activity}   ->  "dashboard › claude -"    broken
{agent} | {activity}   ->  "dashboard › claude |"    broken
{activity} ({agent})   ->  "dashboard › (claude)"    questionable

So the one case the feature exists to serve - a user-chosen separator - is exactly the case where an empty activity leaves garbage in the tab bar. And a silent agent is not an edge case: docs/architecture/title-resolution.md:98-102 describes a session that lived its whole life as 3 · claude.

With the template parsed into prefix / {agent} / middle / {activity} / suffix, Apply knows the literals around each placeholder and drops the ones that surrounded an empty field, instead of hoping Sanitize will. The comment that currently promises more than it delivers then goes away too.

Worth a test: a silent agent under a format whose separator is not ›.

README.md
```diff
@@ -124,6 +124,7 @@ restarts**, with the same `herdr server stop` the install needs.
 | `HERDR_AUTO_TITLE_POSITION`    | `true`                                    | Put each tab's position in front of its title                              |
 | `HERDR_AUTO_TITLE_MANUAL_FILE` | `manual-names.json`, next to `config.env` | Where tabs you renamed by hand are remembered; empty keeps them in memory  |
 | `HERDR_AUTO_TITLE_TRANSCRIPT`  | `true`                                    | Read an agent's own session transcript when it has not titled its terminal |
+| `HERDR_AUTO_TITLE_AGENT_FORMAT` | `{agent} › {activity}`                   | How an agent pane's title reads. `{agent}` is the agent, `{activity}` the work. Drop `{agent}` (say `{activity}`) for a terminal whose own tab HUD already names the agent. Must keep `{activity}` |
```
kryptamine, line 127: **Docs: a word was changed, the decision was not recorded**

- docs/architecture/configuration.md — the only edit is "seven" → "eight". Nothing about the format.
- docs/architecture/title-resolution.md:86-88 still states the rule as fixed: "reappears as a kind, so a tab reads dashboard › claude … and dashboard › claude › Implement OAuth scopes after".
- docs/architecture/title-resolution.md:147 — "qualify binds a kind to whatever a higher source found". For agent panes this is now false: they go through formatAgent, not qualify. The code and the doc disagree in the very paragraph that explains the mechanism.
- README.md:127 — the cell is three sentences, and "for a terminal whose own tab HUD already names the agent" is the why of the feature, which CLAUDE.md places in docs/architecture. The table column padding is also off.

kryptamine (changes requested):

kryptamine:
Thanks for this, and sorry for the slow reply.

Since you opened it, #47 landed with `HERDR_AUTO_TITLE_AGENT_NAME`: set it to `false` and the agent's name stays out of the title. I think that covers what you were after, so I'm going to pass on the format string for now and keep this a plain on/off setting.

Thanks again.

PR #44, opened 2026-09-03
fix(resolver): leave a tunnel-only ssh out of the tab's context

bartekbp:
## Problem

A Claude Code pane whose MCP server opens an `ssh -N -L <port>:db:<port> bastion` port forward gets named `3 · ssh › bastion › claude › <topic>`. Herdr reports the tunnel among the pane's foreground processes (it is a descendant of the agent), and the `ssh` resolver takes the first `ssh` it finds as a remote session.

```
claude
 └─ node <mcp server>
     └─ ssh -N -T -o BatchMode=yes -L 5432:db:5432 bastion
```

## Change

`sshArgs` skips an ssh whose options carry `-N` before the destination, alone or in a cluster (`-NT`, `-fN`). A tunnel runs no remote shell, so the pane's work is still local and the other resolvers name it. A value spelled `N` (`-pN`, `-o N=1`) and a `-N` after the destination (part of the remote command) are not the switch.

Tests cover both sides. `go vet` and `go test -race ./...` pass; `make check` needs Go 1.26 for `tools/`, which I do not have locally.

### DIFF
--- a/internal/resolver/ssh.go
+++ b/internal/resolver/ssh.go
@@ -51,15 +51,49 @@ func (SSH) Resolve(pane *state.PaneState) (Parts, bool) {
 	return Parts{Context: qualify(host, sshKind)}, true
 }
 
-// sshArgs finds an ssh process in the pane and returns its arguments.
+// sshArgs finds an ssh process in the pane and returns its arguments. A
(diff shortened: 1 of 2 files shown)


kryptamine:
Thanks for the contribution! Let’s get this merged.

PR #56, opened 2026-09-07
feat: support windows

### LINKED ISSUE #55: Support Windows
### The problem

Herdr runs natively on Windows (the 0.8.2 preview ships it, and plugins may
claim `windows` in their manifest), but Auto Title does not run there. Three
things stand in the way:

- The manifest lists only `linux` and `macos`, and Herdr enforces that list:
  on Windows it answers `platform_unsupported` and never starts the plugin.
- Even started by hand, the client cannot connect. On Windows
  `HERDR_SOCKET_PATH` (`%APPDATA%\herdr\herdr.sock`) names a 25-byte text file,
  not a socket; Herdr listens on a named pipe instead, so `net.Dial("unix", …)`
  is refused with "the target machine actively refused it".
- The suite cannot say any of this, because every fixture directory is spelled
  `/Users/dev/…`, which `filepath.IsAbs` rejects on Windows, so every tab in the
  tests falls back to `Shell` there (`docs/development.md` records that this is
  why Windows was dropped from CI).

### What it should do instead

The same tab bar as on macOS and Linux, from a checkout that `herdr plugin
install` builds and the Windows server starts:

```
1 · herdr-auto-title › windows
2 · dashboard › claude › Implement OAuth scopes
3 · Windows
```

with the limits Herdr itself imposes on Windows stated in the README rather
than worked around: `pane.process_info` lists only the pane's shell or a
recognized agent there, so `nvim › auth.ts` and `ssh › prod-01` are titles
other platforms get.

Everything below was probed on a live Windows Herdr 0.8.2-preview (protocol
22), not inferred:

- **Transport.** The pipe is `\\.\pipe\` followed by the whole of
  `HERDR_SOCKET_PATH`, drive letter and backslashes included. Framing and the
  one-request-per-connection rule are unchanged. `os.OpenFile` reaches it with
  no new dependency; the handle takes no deadline, but `Close` from another
  goroutine unblocks a pending `Read` on Go 1.24.0 and 1.26.3, which is what
  `context.AfterFunc(ctx, conn.Close)` in the client already relies on.
- **`pane.process_info`.** With `python.exe` and `node.exe` running under a
  pane's `pwsh.exe`, the list held `pwsh.exe` alone; a Claude Code pane listed
  `claude.exe` alone. Names carry `.exe`, a process's `cwd` carries a trailing
  backslash (`C:\github\x\`), and `foreground_cwd` was null on every pane.
  `Set-Location` in pwsh is tracked through Herdr's prompt hook, so the shell's
  `cwd` follows within one prompt.
- **Titles.** An idle pane's `terminal_title_stripped` is Herdr's own
  `pwsh in <dir>` (`pwsh in Windows` after `Set-Location C:\Windows`), which
  says what the context already says.
- **Transcripts.** Claude Code files them under
  `%USERPROFILE%\.claude\projects\<slug>\<id>.jsonl`, same slug rule and same
  `ai-title` lines; the integration hook is PowerShell and reports
  `agent_session.kind = "id"` unchanged.
- **Manifest.** `[[build]]` and `[[startup]]` accept an item-level
  `platforms`, so Windows can build and start `herdr-auto-title.exe` while the
  Unix entries stay as they are.

### What you tried

A branch that makes the fixtures platform-agnostic, dials the pipe on Windows,
strips `.exe`, counts pwsh/PowerShell/cmd as shells, refuses the fallback
title, claims `windows` in the manifest and runs the suite on Windows in CI.
Verified live: linked into a Windows Herdr, the server started
`herdr-auto-title.exe` through `[[startup]]`, every tab was named over the
pipe, a `Set-Location` renamed its tab within a poll, and a manual rename was
left alone. Pull request to follow.

recih:
## What this changes

Auto Title runs on Windows. The one thing that stood in the way was the
transport: on Windows `HERDR_SOCKET_PATH` names a 25-byte text file rather
than a socket, and Herdr listens on the named pipe `\\.\pipe\` + that whole
path, so the Unix dial was refused. The client now opens that pipe on Windows,
with no new dependency, and keeps the Unix dial everywhere else. Around it:
the fixtures take their directories from `herdrtest.Dir` so they are absolute
on every platform, process names lose the `.exe` Windows spells them with,
pwsh/PowerShell/cmd count as shells, Herdr's own idle-pane title
`pwsh in <dir>` is refused as a shell prompt is, drive-rooted paths are
stripped from titles, the manifest claims `windows` with a build and startup
entry of its own, and CI runs the suite on Windows.

Probed on Herdr 0.8.2-preview (protocol 22), recorded in
`docs/architecture/herdr-socket-api.md` and `AGENTS.md`: the pipe name, that
`pane.process_info` on Windows lists only the pane's shell or a recognized
agent (a `python.exe` or `node.exe` child is not listed, so the process and
ssh sources go quiet there), that names carry `.exe` and directories a
trailing backslash, and that `Set-Location` in pwsh is tracked through
Herdr's prompt hook.

Closes #55

## How it was checked

- `go vet`, `go build`, `golangci-lint` and `go test -race` on Windows
  (Go 1.26.3), `go vet`/`go build` on the Go 1.24.0 floor, and `go vet` for
  linux and darwin. `make` is not in Git Bash, so the recipes ran by hand.
- The client tests on Windows talk to a named pipe server of their own, so
  the transport the plugin ships is the one the suite exercises.
- Linked the checkout into a live Windows Herdr: the server started
  `herdr-auto-title.exe` through `[[startup]]`, every tab was named over the
  pipe, a `Set-Location` renamed its tab within a poll, a manual rename was
  left alone, and a Claude Code session with a generic terminal title was
  named from its transcript under `%USERPROFILE%\.claude`.

## Before merging

### DIFF
--- a/internal/app/app_test.go
+++ b/internal/app/app_test.go
@@ -1105,3 +1105,27 @@ func TestRunNamesWhatExistsBeforeTheFirstTick(t *testing.T) {
 	cancel()
 	<-done
 }
+
+func TestAWindowsShellPaneIsNamedAfterItsDirectory(t *testing.T) {
+	// What Herdr reports for an idle pane on Windows: the shell with its
+	// extension, its directory with a trailing separator, and a title of
+	// Herdr's own making that names both. None of it is what the pane is doing.
+	h := start(
+		t,
+		[]herdr.TabInfo{{TabID: "wE:t1", Label: "1"}},
+		[]herdr.PaneInfo{{
+			PaneID: "wE:p1", TabID: "wE:t1", Focused: true,
+			CWD:                   dashboard,
+			TerminalTitleStripped: "pwsh in dashboard",
+		}},
+	)
+	h.client.SetProcesses("wE:p1", herdr.PaneProcessInfoProcess{
+		Name: "pwsh.exe",
+		CWD:  dashboard + string(filepath.Separator),
+	})
+	h.poll()
+
+	if got := h.client.Renames()[0].Label; got != "dashboard" {
+		t.Errorf("rename = %q, want dashboard", got)
+	}
+}
--- a/internal/app/app_test.go
+++ b/internal/app/app_test.go
@@ -66,10 +66,6 @@ type harness struct {
 	t      *testing.T
 	app    *App
 	client *herdrtest.Client
-	// polled is when the last poll finished. The clock is what orders the
-	// changes a poll sees, and on Windows it ticks too coarsely to tell two
-	// polls apart unless one waits for it.
-	polled time.Time
 }
 
 func start(t *testing.T, tabs []herdr.TabInfo, panes []herdr.PaneInfo) *harness {
@@ -89,12 +85,7 @@ func startConfigured(t *testing.T, client *herdrtest.Client, cfg Config) *harnes
 func (h *harness) poll() {
 	h.t.Helper()
 
-	for !time.Now().After(h.polled) {
-		time.Sleep(time.Millisecond)
-	}
-
 	h.app.poll(context.Background(), h.client)
-	h.polled = time.Now()
 }
 
 func (h *harness) polls(n int) {
@@ -102,6 +93,16 @@ func (h *harness) polls(n int) {
 
 	for range n {
 		h.poll()
+	}
+}
+
+// awaitClock returns once the wall clock has moved on. Which pane changed last
+// is told by the clock, and on Windows it ticks too coarsely to tell two polls
+// apart unless one waits for it.
+func awaitClock() {
+	start := time.Now()
+	for !time.Now().After(start) {
+		time.Sleep(time.Millisecond)
 	}
 }
 
@@ -347,6 +348,7 @@ func TestTheMostRecentlyChangedPaneNamesTheTab(t *testing.T) {
 	h.client.SetPane(herdr.PaneInfo{
 		PaneID: "wE:p2", TabID: "wE:t1", Revision: 2, CWD: api,
 	})
+	awaitClock()
 	h.poll()
 
 	if got := h.client.Renames()[1].Label; got != "api" {
--- a/internal/app/app_test.go
+++ b/internal/app/app_test.go
@@ -19,6 +19,14 @@ import (
 
 const testPoll = 10 * time.Millisecond
 
+// The directories the fixtures sit in, absolute on whichever platform the
+// tests run on: a relative directory names no tab, and Windows has no /Users.
+var (
+	dashboard = herdrtest.Dir("work", "dashboard")
+	api       = herdrtest.Dir("work", "api")
+	billing   = herdrtest.Dir("work", "billing")
+)
+
 func testConfig() Config {
 	return Config{
 		Poll:      testPoll,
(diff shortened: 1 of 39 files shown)


internal/claude/transcript.go
```diff
@@ -190,8 +190,11 @@ func (r *Reader) locate(sessionID, dir string) (string, bool) {
 }
 
 // slugOf is how Claude Code names a project directory: every character that is
-// not a letter or a digit becomes a dash, the leading separator included.
+// not a letter or a digit becomes a dash, the leading separator included. A
+// trailing one is not, because the directory Claude Code names carries none.
 func slugOf(dir string) string {
+	dir = filepath.Clean(dir)
```
kryptamine: Could this cleanup move up to state.PaneDir / snapshotDir instead? claude knows how Claude Code names a project, and it would be nice if it didn't also have to know how Herdr spells a directory on Windows.
Right now three readers of PaneState.Dir each clean it on their own: git.discover, CWD.base, and now slugOf. If the directory were cleaned where it enters the domain, next to programName which already strips .exe at that same boundary, none of the three would have to think about it.
One catch: filepath.Clean("") returns ".", and an empty Dir is meaningful (there's a test for the fallback), so it would need an early return for the empty case.

internal/app/app_test.go
```diff
@@ -68,7 +88,13 @@ func startConfigured(t *testing.T, client *herdrtest.Client, cfg Config) *harnes
 // exercises what the loop does rather than a shortcut past it.
 func (h *harness) poll() {
 	h.t.Helper()
+
+	for !time.Now().After(h.polled) {
```
kryptamine: This makes every test in the package wait on the wall clock, while only one or two of them actually depend on the order of changes across polls (`TestTheMostRecentlyChangedPaneNamesTheTab` is the one I can find).

Worth considering: `ChangedAt` is never compared against a duration. Its only use is `p.ChangedAt.After(best.ChangedAt)` in `state/tab.go`, for picking the pane that moved last. A poll counter in Changes, incremented in Observe, would order panes the same way, would be exact on every platform, and this wait could go. Real time would stay where it's genuinely needed, in `readAt` and `processRefresh`.

That's a separate logical change, so a preceding refactor PR would be cleaner than folding it in here. If you'd rather keep the wait for now, could it sit in the two tests that need distinct timestamps instead of in the shared poll()?

kryptamine (changes requested):

internal/app/app_test.go
recih: Done in e3b7688. The wait now sits in `TestTheMostRecentlyChangedPaneNamesTheTab` alone, as `awaitClock()` between its two polls, and the shared `poll()` no longer touches the clock. With the wait removed altogether, that test was the only one in the package to fail on Windows over forty runs of the suite, so it is the only one that gets it.

Agreed that a poll counter in `Changes` is the right ordering: `ChangedAt` is only ever compared, never measured. I would rather land that as a refactor of its own once this is in than widen this PR.

internal/claude/transcript.go
recih: Done in 1cb415a. `snapshotDir` and `PaneDir` now pass what Herdr reports through `cleanDir`, next to `programName`, and `slugOf` is back to what it is on main. `cleanDir` keeps `""` as `""` for the reason you gave, and the new state test covers the snapshot's directory, a process read's, and the empty one. Two existing state tests had to take their directories from `herdrtest.Dir`, since on Windows `filepath.Clean` respells `/work/x` with backslashes.

The Cleans in `git.discover` and `CWD.base` predate this PR and are now redundant. I left them alone to keep the PR to Windows; happy to drop them here if you would rather.

kryptamine (approved):

recih:
Added 734b804: `scripts/probe.py` decoded Herdr's output and printed in the locale's code page, which on Windows is not UTF-8, so the first label with `›` or non-Latin text broke it. It now reads and prints UTF-8 explicitly. Verified with `tabs`, `snapshot` and `version` on Windows, piped and not.

PR #58, opened 2026-09-08
fix: leave when another server takes the socket

### LINKED ISSUE #57: The plugin outlives the server that started it and runs beside the next one's
### What happened

After the Herdr server was restarted, two `herdr-auto-title` processes were
polling the same socket: the one the previous server had started through the
startup hook, still alive with its parent gone, and the one the new server
started. Both renamed tabs. The two agree only as long as they run the same
settings; once they differ — the old one keeps whatever `config.env` said when
it started — each takes the other's rename for the user's and locks the tab,
because the manual rename protection has nothing else to go by.

### What you expected instead

One instance per server. When the server that started an instance is gone and
another one holds the socket, that instance should leave: the new server has
already started one of its own.

### How to reproduce it

1. Run Herdr with Auto Title installed and note the plugin's pid.
2. `herdr server stop`, then bring Herdr back.
3. List processes: the old pid is still polling, beside the new one. The socket
   path is the same, so both connect to the new server.
4. Change any setting in `config.env` between the two starts and the two
   instances disagree on titles from then on.

### Where it comes from

In Herdr (`herdrdev/herdr` at 9e01168), `start_plugin_command` in
`src/app/api/plugins/runtime.rs` spawns a startup hook and waits on it in a
thread of its own, keeping no handle; `complete_shutdown` in
`src/server/headless/lifecycle.rs` closes clients and removes the socket files
and touches no plugin process; and `src/server/headless/bootstrap.rs` calls
`run_plugin_startup_hooks` from both of its entry points, a fresh start and a
live handoff import. Herdr's plugin docs call startup hooks "one-shot
initialization commands rather than supervised daemons", so a daemon has to
notice on its own that its server is gone — and nothing in Auto Title does:
`poll` never gives up, by design, and `main.go` exits only on `SIGINT` and
`SIGTERM`. A live handoff (`herdr update --handoff`) runs the hooks again the
same way, so every platform is affected.

### Auto Title version

0.4.0, a linked checkout of `main` at a34f22d.

### Herdr version

herdr 0.9.0-preview.2026-09-08-62431dbd033b; the source read is 9e01168.

### Operating system

Windows, where it was seen; the code paths above are the same on macOS and
Linux.

### What you tried

Each instance remembers which server it answers to and leaves when another
takes the socket. Herdr itself tells its own socket from a successor's by the
socket file (`socket_file_identity` in `src/ipc.rs`): the file's device and
inode on macOS and Linux, the `<pid>:<start ns>` marker written into it on
Windows. A `Client.Server` reading the same thing costs one stat per poll; the
loop records the first identity it can read and ends the run on the first poll
that reads a different one, with exit status 0, since the successor's startup
hook has already started the replacement. While no server can be read nothing
is decided, so an outage or a slow start is still just a run of failed polls.
Verified in a throwaway session: the old build stayed after a server restart,
the fixed build logged that another server held the socket and left within a
poll. Pull request to follow.

recih:
## What this changes

An instance of Auto Title now leaves when another server takes the socket.
Herdr starts a plugin through a one-shot startup hook and keeps no handle on
it, so a server that stops leaves the process running and the next server
starts a process of its own, at a fresh start and after a live handoff alike;
the two then rename against each other, and once their settings differ each
takes the other's rename for the user's and locks the tab. `Client.Server`
reads the socket's identity the way Herdr tells its own socket from a
successor's (`socket_file_identity` in `src/ipc.rs`): the socket file's device
and inode on macOS and Linux, the `<pid>:<start ns>` marker Herdr writes into
the file on Windows. The loop records the first identity a poll can read and
ends the run on the first poll that reads a different one, exit status 0, since
the successor's startup hook has already started the replacement. An
unreadable identity decides nothing, so an outage or a slow start is still a
run of failed polls, as before. The Herdr code paths this rests on are cited
in `docs/architecture/poll-loop.md` and `herdr-socket-api.md`.

Closes #57

## How it was checked

- Rebased onto `main` after #56 landed: `gofmt`, `go vet`, `golangci-lint`
  and the whole suite under `go test -race` on Windows, plus `go vet` for
  linux and darwin and on the Go 1.24.0 floor. The fork's CI ran the same
  branch on every job, Windows included: https://github.com/recih/herdr-auto-title/actions/runs/34199495704
- The identity read is tested against a real socket on macOS and Linux — a
  rebind at the same path, with the old file kept aside so a recycled inode
  cannot fake a match — and against the marker file on Windows.
- Live, in a throwaway session (`herdr --session at-probe server`): the
  previous build stayed after the server was stopped and started again; this
  build kept polling through the outage and, once the new server was up,
  logged that another server held the socket and left within one poll.

## Before merging

      Windows ships no `make`).

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -174,6 +174,13 @@ are only the facts that would otherwise mislead the code in silence.
   shell of whoever installed it, which is why `HERDR_AUTO_TITLE_*` settings
   arrive through `config.env` (see
   [docs/architecture/configuration.md](docs/architecture/configuration.md)).
+- **Herdr keeps no handle on a plugin it started.** A startup hook is spawned
+  and forgotten: the server stops nothing when it stops, and runs the hooks
+  again at every start and live handoff, so an instance that stayed would run
+  beside its successor and lock the tabs the two name differently.
+  `App.superseded` is what makes it leave instead; the loop must not survive
+  a change of server (see
+  [docs/architecture/poll-loop.md](docs/architecture/poll-loop.md)).
 
 ## Working here
 
--- a/README.md
+++ b/README.md
@@ -41,7 +41,8 @@ herdr plugin install kryptamine/herdr-auto-title
 
 Herdr clones the repository, builds the binary and registers it. Until the
 server has restarted, nothing is renamed. `herdr plugin list` shows the plugin,
-`herdr plugin disable` turns it off.
+`herdr plugin disable` turns it off. Every server start brings a fresh instance,
+and the one the previous server started leaves on its own.
 
 ### Better names for agent tabs
 
--- a/docs/architecture/herdr-socket-api.md
+++ b/docs/architecture/herdr-socket-api.md
@@ -64,6 +64,20 @@ exists here.
 > `herdr server stop`, which closes the session. The directory and the two CLI
 > commands were confirmed directly. Auto Title depends on none of it.
 
+**A startup hook is run and forgotten.** In the Herdr source,
+`start_plugin_command` (`src/app/api/plugins/runtime.rs`) spawns the command
+and waits on it in a thread of its own, keeping no handle; server shutdown
+(`complete_shutdown`, `src/server/headless/lifecycle.rs`) closes clients and
+removes the socket files and touches no plugin process; and
+`run_plugin_startup_hooks` runs from both server entry points
+(`src/server/headless/bootstrap.rs`), a fresh start and a live handoff import.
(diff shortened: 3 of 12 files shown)


recih:
Rebased onto main now that #56 has landed (both changed the test harness's `poll`); the one commit is c34845c. With the Windows support underneath it the whole suite runs on Windows too, and the fork's CI is green on every job including the Windows one: https://github.com/recih/herdr-auto-title/actions/runs/34199495704

kryptamine (approved):

PR #60, opened 2026-09-09
feat: HERDR_AUTO_TITLE_PREFER_AGENT names a split tab after its agent pane

youngxguo:
## Problem

A tab holding an agent and an editor is named after whichever pane is focused. Split nvim beside claude and the tab flips between `nvim › file.ts` and the agent's topic on every focus change, which is noisy in the tab bar and in Herdr's agent sidebar (where the `tab` token repeats it).

## Change

`HERDR_AUTO_TITLE_PREFER_AGENT=true` puts one rule in front of the existing selection: a tab holding a pane that Herdr recognizes an agent in is named after that pane, whichever pane is focused. Idle agents count too, since the point is a stable name. Off by default, so nothing changes for anyone who has not asked for it.

- `state.SelectContextPaneWith(tab, preferAgent)` carries the rule; `SelectContextPane` keeps its signature and behavior.
- `resolver.Options.PreferAgentPane` and `Config.PreferAgentPane` plumb it through, mirroring `HideAgentName`.
- Documented in the README table and `config.env.example`.

## Testing

`go test ./...` passes; `TestSelectContextPaneWithPreferAgentOutranksFocus` covers both settings on the same tab. Running it here with claude in one pane and nvim focused in the other keeps the tab on the agent's title.

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,17 +76,18 @@ the path for your platform and uncomment what you need:
 Auto Title reads the file once at startup, so run `herdr server stop` after a
 change. It does not read the config directory that `herdr plugin list` prints.
 
-| Setting                        | Default                                  | What it does                                                      |
-| ------------------------------ | ---------------------------------------- | ----------------------------------------------------------------- |
-| `HERDR_AUTO_TITLE_DEBUG`       | `false`                                  | Log at DEBUG instead of INFO                                      |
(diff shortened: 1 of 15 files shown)


kryptamine:
Thanks for the contribution! I'll check it over the weekend.

PR #61, opened 2026-09-09
fix(resolver): do not repeat a branch its worktree is named after

xiaoyu2er:
## The problem

`git worktree add ../feat-oauth feat-oauth` makes a directory and a branch of the same name. `withoutRepetition` compares the activity against the context and against the branch, but never the context against the branch, so the title reads:

```
feat-oauth › feat-oauth › nvim
```

A third of the tab bar spent on the word it has just said. Reproduced against a live session of six worktrees — every one of them was affected:

```
before   1 · pane-rename › pane-rename › herdr-au
after    1 · pane-rename › herdr-auto-title pane

before   1 · fix-dev › fix-tb-dev
after    1 · fix-tb-dev
```

## The fix

The directory leads a title and the branch only qualifies it, so the branch is the half that goes.

**The match is exact**, as the workspace rule below it is. A worktree whose directory spells its branch differently (`xl-knp-3` against `xl-knp.3`) keeps both, because nothing here can tell a near-miss from two real facts. That is the same trade the workspace rule already documents, so this does not widen it.

## Checks

`make check` is green — fmt, vet, golangci-lint (0 issues) and `go test -race`.

The new test fails without the change, with `pane-rename › pane-rename › claude › herdr-auto-ti`.

### DIFF
--- a/docs/architecture/title-resolution.md
+++ b/docs/architecture/title-resolution.md
@@ -279,6 +279,18 @@ is how Windows spells one, and Herdr's own title for an idle pane there,
 `pwsh in dashboard`, is refused as a shell prompt is: it names the shell and
 the directory the context already names.
 
+## A worktree does not say its branch twice
+
+`git worktree add ../feat-oauth feat-oauth` makes a directory and a branch of
+the same name, and the two are one fact rather than two: the title would read
+`feat-oauth › feat-oauth › nvim`. The branch is dropped when it matches the
+directory exactly, because the directory leads the title and the branch only
+qualifies it.
+
+Exactly, and no more than that — the rule below makes the same trade. A worktree
+whose directory spells its branch differently (`xl-knp-3` against `xl-knp.3`)
+keeps both, because nothing here can tell a near-miss from two real facts.
+
 ## The workspace is not repeated
 
 Herdr shows the workspace above its tabs, so a tab in the workspace it is named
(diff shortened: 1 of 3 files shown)


kryptamine:
Thanks for the contribution! I'll check it right away.

internal/resolver/resolver_test.go
```diff
@@ -322,3 +323,21 @@ func TestTheHomeDirectoryIsMatchedTheWayWindowsSpellsIt(t *testing.T) {
 		t.Errorf("name = %q, want %q", got.Name, GenericFallback)
 	}
 }
+
+func TestAWorktreeNamedAfterItsBranchSaysItOnce(t *testing.T) {
```
kryptamine: Could you please move this test to `git_test.go`?

kryptamine (changes requested):

internal/resolver/resolver_test.go
xiaoyu2er: Moved to `git_test.go`.

It uses `repoPane` and `resolveRepoPane` now, so the collision is expressed with the fixtures already in that file — `repoPane("dashboard", "main")` sits in the `dashboard` directory on a branch of the same name, which is the shape `git worktree add ../dashboard dashboard` produces.

Rechecked that it still fails without the change: `title "dashboard › dashboard › nvim › auth.ts", want the branch said once`.

kryptamine (approved):

PR #62, opened 2026-09-09
feat(app): name panes as well as tabs

xiaoyu2er:
## Why

Herdr's goto panel (`prefix`+`g`) lists a pane by `pane.label`, falling back through the agent's name, its display name, its title and finally `pane N`. Auto Title only ever called `tab.rename`, so a session of Claude Code panes reads as a column of `claude` with nothing to pick between them.

## What it does

`HERDR_AUTO_TITLE_PANES=true` names each pane with the chain that names tabs, from that pane's own state rather than from the pane its tab speaks through.

A pane's row sits **under** its tab's, so the parts the tab already carries are dropped and the activity is kept — where a pane is has a row of its own above it, and what it is doing is the whole of what a pane row is for:

```
pane-rename › herdr-auto-title pane      the tab, truncated
  herdr-auto-title pane 重命名            the pane the tab speaks through
  herdr-reviewr                          the pane doing something else
```

The first row ends up carrying **more** than its tab, which had to truncate the same words to fit the directory in front of them.

A pane whose only fact is its directory does repeat its tab. There is nothing else known about such a pane, and the alternative is the `claude` this setting exists to replace.

## Manual rename protection

Extended to panes, and this is the part that matters most: without it a name you typed would be overwritten twice a second.

`Manual` now holds a set of claims per kind. Locks for both live in the same file, panes under `locked_panes`, so a store written by an older version reads back unchanged. **The two locks are independent** — a tab named by hand says what the tab bar should read and says nothing about how the panes inside it should be listed.

## Probed, not assumed

Two facts about `pane.rename` differ from `tab.rename` and both are recorded in `docs/architecture/herdr-socket-api.md`:

- **A pane's label is absent from the wire until the pane has one.** Across a seven-pane session no pane object carried `label` in `session.snapshot`, `pane.get` or `pane.list`; renaming one made the key appear in all three, and clearing it made the key vanish again.
- **`pane.rename` clears an empty label rather than storing it.** Both `{"pane_id": p}` with no label and `{"pane_id": p, "label": ""}` answered with the key gone. So a pane has **one** unnamed spelling where a tab has two, and there is no position spelling to accept beside it.

## Why it is off by default

It changes what an existing user sees, and it costs a `pane.process_info` read per pane rather than per tab — a cost that scales with how the user splits. A test pins that number so a change would move it.

## Checks

`make check` is green — fmt, vet, golangci-lint (0 issues) and `go test -race`.

Verified against a live 21-pane session. The protection tests were checked by removing the protection: `TestAPaneTheUserRenamedIsLeftAlone` then fails with the pane's manual name overwritten, which is the failure it exists to catch.

Not verified: Windows. Only macOS hardware was available.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -143,8 +143,12 @@ are only the facts that would otherwise mislead the code in silence.
   agent**, never an editor, a build or an ssh session running under the shell.
   Names arrive with `.exe` and a process's `cwd` with a trailing backslash; the
   state package strips both as they arrive, so no reader of a pane sees either.
-- Auto Title uses three methods and no others: `session.snapshot`,
-  `pane.process_info` and `tab.rename`.
+- Auto Title uses four methods and no others: `session.snapshot`,
+  `pane.process_info`, `tab.rename` and `pane.rename`.
+- **A pane carries no label until it has one, and an empty one clears it.**
+  Herdr omits `label` from a pane object entirely until the pane is named, and
+  `pane.rename` clears rather than stores an empty label — the opposite of
+  `tab.rename`. So a pane has one unnamed spelling and a tab has two.
 - **Do not reintroduce an event subscription.** `events.subscribe` replays
   about ten seconds of backlog per pane before anything live, with no cursor to
   skip it, so a subscriber opens by reacting to a session that is gone. A
--- a/README.md
+++ b/README.md
@@ -71,6 +71,31 @@ saying it twice: `HERDR_AUTO_TITLE_AGENT_NAME=false` leaves it out, and
 where the name was all there was — a pane whose agent has reported nothing then
 reads as its directory, `dashboard`, in place of `claude`.
 
+## Naming panes as well as tabs
+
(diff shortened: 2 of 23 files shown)


kryptamine:
Thanks for the contribution again! I'll take a careful look at it over the weekend and make sure I fully understand the feature and the proposed change.

kryptamine:
@xiaoyu2er 

Thanks for this, it's a great feature! I pushed a few commits on top of your branch:

- rebased it onto main
- merged tab and pane naming into one code path, which also fixed a bug where a pane could get a different name once its tab was renamed by hand
- turned pane naming on by default

The feature is worth far more than it costs. A `pane.process_info` read takes about 0.17 ms, and a pane reuses its last answer for up to 2 seconds, so each pane costs somewhere between one read every 2 seconds when idle and two reads a second when busy. Even twenty busy panes at once add up to roughly 7 ms of Herdr's time per second, under 1% of a core.

It does change what existing users see, so the switch is marked as a breaking change, and `HERDR_AUTO_TITLE_PANES=false` brings back the old behaviour.

PR #65, opened 2026-09-10
feat: slide a title too wide for the tab bar

### LINKED ISSUE #64: A title too wide for the tab bar loses its tail
### The problem

At the default fifty columns a title rarely overflows, but the narrower the
tab bar the more often the tail goes — and the tail is the part that says what
the tab is doing. At twenty columns, `dashboard › claude › Implement OAuth
scopes` renders as `dashboard › claude`. Three tabs holding three projects with
an agent in each then read alike, and the one thing that told them apart is
what was cut.

Korean and Japanese titles reach it sooner still: each character takes two
columns, so twenty columns is ten characters.

### What it should do instead

A title too wide for the bar should move through the tab, a window at a time,
so all of it is readable over a few seconds:

    dashboard › claude
    ashboard › claude ›
    shboard › claude › I
    ...
    mplement OAuth scope

It should open on the head and hold there long enough to read where the title
starts, then come back round to the head after a gap. Off by default — a tab
bar that holds still is what most people want, and a title that fits should
never move.

Nowhere new: it is the title the resolver already produces. What is new is that
the fit stops being a single cut. Which part of the title shows depends on how
many polls have passed, so the loop has to carry that per tab, the way it
already carries what each pane was running.

Two things worth settling here rather than in review:

- **Speed.** One column per poll at the default 500 ms means a 44-column title
  takes 51 polls, about 25 seconds, for one lap. A step measured in columns per
  poll would tune that without lowering `POLL_MS`, which speeds up every read
  the loop makes for every tab.
- **The position.** It is the key that switches to the tab, so it should stay
  pinned while the rest slides past it.

### What you tried

Lowering `MAX_LENGTH` cuts sooner rather than helping. Raising it past the tab
bar's real width does not help either: Herdr sizes a tab to its label, so one
long title takes room from every other tab. Renaming the tab by hand turns Auto
Title off for it, which is the wrong trade for a tab whose work keeps changing.

aimdevlee:
## What this changes

**_Tab section_**

https://github.com/user-attachments/assets/7225e4f2-641c-430b-b934-3643865fac77

**_Agents section_**

https://github.com/user-attachments/assets/57f196f2-f0ab-40a4-bebf-f47401d6d931

A title wider than `MAX_LENGTH` is cut, and the tail is where a tab says what it is doing: at twenty columns `dashboard › claude › Implement OAuth scopes` becomes `dashboard › claude`. `HERDR_AUTO_TITLE_SCROLL` shows a window of the title instead — the head held still for four polls, then moved `HERDR_AUTO_TITLE_SCROLL_STEP` columns per poll, round through a three-column gap to the head again. The step is the speed knob and the poll interval is deliberately not: moving further per poll costs nothing, while lowering `POLL_MS` speeds up every read the loop makes for every tab in the session.
Both settings default to what happens today, so nothing changes for anyone who does not ask for it.

The first commit is the prerequisite: the chain bounded a title and the position decorator bounded it again. Cutting twice was harmless — the same cut, one column further in — but sliding twice advances the window twice per poll, so the fit had to become one thing in one place.

Closes #64 

## How it was checked

- `make check` is green.
- Run in a live session at `MAX_LENGTH=20`, `SCROLL_STEP=2`, across Korean, Japanese and ASCII titles.
- A first live run jittered: the tab grew and shrank by a column as the window slid through two-column characters, because a cluster that straddles the edge is dropped whole. Measured over 60 frames of one real title, 11 came out at 19 columns; every window is now padded to the width it was given and the count is 0.
- The renames a slide makes are the plugin's own, so a tab whose label changes every poll must not read as one the user renamed. Covered by a test, as is a tab id Herdr hands out again starting at the head rather than mid-lap.
- The position stays put while the title slides past it. That guarantee was stated in the docs and unverified; it has a test now.

## Before merging

- [x] One logical change per commit, Conventional Commits, no co-author trailer. **PRs are rebased, never squashed or merged**, so the commit messages are what lands in the history and in the changelog.
- [x] `CHANGELOG.md`, the tags and the version in `herdr-plugin.toml` are untouched — they belong to release-please.
- [x] A probe that taught something new is recorded in `AGENTS.md` and in `docs/architecture/herdr-socket-api.md`.


kryptamine:
Thanks for the contribution! I'll check it over the weekend.

kryptamine:
Could you please rebase this onto `main` and resolve the merge conflicts? Thank you!

aimdevlee:
@kryptamine 
Rebased onto `main`. Two notes on how I resolved it, and one question.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                    
                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                          
- **`Fitted` (was `Numbered`) implements `PaneResolver` too.** With the chain left unbounded, returning the bare chain as the pane resolver would have dropped the width bound on pane labels — `TestTheSettingsThatShapeATitleShapeAPaneLabel` ("the length limit binds a pane too") is what pinned this down for me.                                                                                                                                                                                                                                                                                                                                                                    
- **Sliding therefore applies to pane labels as well**, keyed by pane id, which seemed to follow from that test's premise that everything tuned about a title holds for a pane. Added `TestAPaneLabelSlidesToo` to cover it.                                                                                                                                                                                                                                                                                                                                                                                                                                                              
                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                          
The question: **`Resolvers` is now an unexported `resolvers(cfg, fit)` called from `App.New`.** The reason is that the scroller is state carried between polls — pruned in `Retain`, read in `apply` so a slide logs at DEBUG rather than as a rename — so it belongs to the `App`, while the resolvers need its `Fit`. Your function survives with four lines changed, but its visibility and call site moved, and I'd rather that be your call than mine. Happy to restore an exported entry point if you prefer — e.g. `Resolvers(cfg) *Naming` returning the two resolvers plus the scroller behind an unexported field, so `main` keeps doing the wiring. Just say which you'd like. 
                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                          
Two small doc fixes came along for the ride: `configuration.md` counted nine `HERDR_AUTO_TITLE_*` variables where there were ten (twelve now), and still pointed at `main.run` for where they are consumed, which has been `app.New` since the pane-naming refactor.

kryptamine:
@aimdevlee 
Hi! Sorry for the slow feedback, and thanks for the careful work here. I've taken a close look at the problem and the proposed solution.

The problem in #64 is real, but I'm not sure scrolling is the right fix:

- People scan a tab bar. With scrolling you have to wait for the part you need, and a 44-column title takes about 25 seconds for one lap.
- Several titles moving at once are distracting while you work in a pane.
- A sliding tab is renamed up to twice a second, which in your measurement came to 236 KB of Herdr's server log a minute.
- `Fit` gives the resolver state, so the same session no longer resolves to the same title.

On CJK, a two-column character also carries more meaning than a Latin letter, so I see it as the same narrow-bar problem.

A few settings already make titles shorter, so more of the tail fits:

- `HERDR_AUTO_TITLE_AGENT_NAME=false` drops the agent's name, so `dashboard › claude › Implement OAuth scopes` becomes `dashboard › Implement OAuth scopes`.
- `HERDR_AUTO_TITLE_BRANCH_MAX` limits how many columns the branch takes; `0` leaves it out.
- `HERDR_AUTO_TITLE_POSITION=false` drops the `1 · ` in front, which saves four columns.

As I see it, the actual bug is what the cut drops: it keeps the project and agent and loses the activity, which is what tells tabs apart. That could be fixed without motion by shortening context and branch before the activity (`dash… › Implement OAuth…` instead of `dashboard › claude`) and marking the cut with an ellipsis.

The first commit, cutting a title in one place, looks worth having either way.

What do you think? Would these settings cover your case, or is there something scrolling gives you that a smarter cut wouldn't?

aimdevlee:
@kryptamine 
The settings you listed are enough for what I actually needed, so scrolling is paying a real price for something two flags already give me. Your fourth point settles it: the package promises that identical state yields an identical decision, and an opt-in that quietly breaks that is not worth the tail of a title.

So I'm dropping the scrolling and closing this.


Thanks for taking the time :pray:

Issue #69, opened 2026-09-14
Name the workspace as well as its tabs and panes


chechunhsu:
### The problem

In my session the workspace row keeps the label it was created with while every tab under it has moved on.

Auto Title already reads that label on every poll and has an opinion about it — `title-resolution.md` drops it from a tab title when the tab would only repeat it — but never writes it.

### What it should do instead

The workspace follows the worktree its tabs are in, behind a default-off setting in the shape `HERDR_AUTO_TITLE_PANES` already has:

```
dashboard › feat/oauth
```

Three things I cannot settle from outside:

- `MAX_LENGTH` is 50 columns of the tab bar. The workspace label renders somewhere else, so does it want a budget of its own, and what default?
- Whether a branch belongs there at all. If it does, a tab in that workspace may want to stop carrying the same one, which is the dedup rule's business.
- Which tab speaks for the workspace when its tabs disagree — the active one, only what every tab has in common, or nothing. `TabFrom` answers this a level down; there is no cross-tab rule.

### Where the plugin would read it from

`tabsIn` already maps `workspace_id → label`, so that read is paid for. The rest is not: the Go `WorkspaceInfo` models only the id and label it reads today, and the wire object's worktree fields are `repo_key`, `repo_name`, `repo_root`, `checkout_path` and `is_linked_worktree` — no branch, so a branch would come from `internal/git` against the checkout path.

The schema exposes two write paths:

- **`workspace.rename`**, taking `{workspace_id, label}`, the same shape as `tab.rename`. It sets the label, so it inherits the manual rename problem whole.
- **`workspace.report_metadata`**, taking `{workspace_id, source, tokens}` and an optional `ttl_ms`. The CLI calls it display-only, it is attributed to a source, and it never sets the label. That does not retire the manual rename question so much as change it: if a user renames a workspace by hand, should source-attributed metadata keep changing that row beside it? I have not found anything that says how, or whether, the sidebar renders those tokens. You probed the pane half and recorded that nothing installs a source for it; the workspace half is not in the repo.

Worth probing either way: `WorkspaceInfo.number` exists, and reading `TabInfo.number` as the default label is what locked every tab created after the first poll. Whether an unnamed workspace shows its position, and whether naming one costs the key that jumps to it, decides if the sidebar needs numbering of its own — `Numbered` takes a `TabState` and is tab-specific.

### What you tried

Renaming by hand, which is the thing that does not survive the work moving.

6temes:
[...]

PR #70, opened 2026-09-14
feat: name a one-tab workspace's row after that tab

chechunhsu:
Implements #69 as settled there.

## What it does

`HERDR_AUTO_TITLE_WORKSPACES=true` names a workspace holding **exactly one tab** after that tab, under `HERDR_AUTO_TITLE_WORKSPACE_MAX_LENGTH` (default `20`). Off by default.

## Behaviour

Run against the stub session, not written from the spec — these are the labels the tests actually observe.

| The workspace | The row becomes | Why |
|---|---|---|
| one tab, repo on `feat/oauth`, label still `api` | `api › feat/oauth` | the ordinary case |
| one tab, repo on the default branch, no agent | *left alone* | the derived name is the label it already has, so there is nothing to set |
| one tab, agent reporting `Fix invoice totals` | `Fix invoice totals` | context and agent name dropped to fit 20 — your worked example, reproduced exactly |
| one tab, branch **and** agent topic | `Fix invoice totals` | same; at 20 columns a topic of ordinary length is the whole row |
| one tab, `nvim` with the terminal title `auth.ts` | `dashboard › auth.ts` | the title is read without the program that set it; the tab under it still says `nvim › auth.ts` |
| label is `the migration` | *left alone* | not the directory's basename, so it is a name its owner wrote |
| two tabs | *left alone* | no cross-tab rule to apply |
| tab holds no pane, or its pane sits at `/` | *left alone* | no basename to judge the label by; judged on the first poll that shows one |
| opened after startup, labelled after its directory | named as any other | not claimed merely for being unseen |
| named by this, then the plugin restarts | named on | the label last written is kept in the lock file, so its own work is not read as yours |
| `npm run build` running in the pane | unchanged by it | the foreground process is the one source the row's chain leaves out; the tab still takes it |

Truncation drops whole parts from the front and cuts only what survives, from the end:

```
demo-repo › feat/billing › Fix invoice totals   45
feat/billing › Fix invoice totals               33
Fix invoice totals                              18  ← set
```

Tabs keep `MAX_LENGTH`'s 50 and cut from the end, unchanged.

## What I could not settle

**The identity test cannot see what it is testing against.** The rule is that a workspace still wearing its directory's basename is unnamed. The snapshot carries no creation directory, so that basename is read from the pane its tab is in — and those are the same thing only until someone `cd`s. A workspace made in `~/work/api` whose tab has moved to `~/work/web` reports `web` as its default, so the genuine default `api` reads as a name its owner wrote and the row is never taken; name it `web` deliberately and it reads as Herdr's own and is.

That is what the rule as stated does, so I have implemented it as stated.

**The row takes the terminal title, without the program that set it.** You listed the row as context, branch and agent topic. The chain also carries the terminal title, which for Claude Code *is* the topic, but for `nvim` is the file it has open. The row takes it and drops the program name — `auth.ts` where the tab says `nvim › auth.ts` — so it does not change when the program does. Kept on @6temes's review: a pane running an editor would otherwise have nothing but its directory.

A workspace with no basename to judge by — its tab has no pane yet, or the pane sits at `/` — is neither claimed nor named until a poll shows one; claiming it would lock it under Herdr's own label, and naming it would write over the owner's. And the row is named after the tabs in the same poll, so for one poll after it is first named the tab under it still repeats the branch.

**Two places a tab still repeats the row.** It drops any of its own parts the row carries, but only those it can match:

- the *activity*, which the existing rule never drops because it is what a row is for. Under one tab the row and the tab resolve from the same pane, so it always repeats. Dropping it is a two-line change confined to the row comparison — it does not touch how a pane reads its tab, which is what I first assumed — so this is a decision rather than a constraint.
- a part the row had to **cut**. With a twenty-column bound and a long project name the row reads `a-very-l` while the tab's part is still `a-very-long-project-name`; they no longer compare equal, so the tab says it again. Prefix matching would fix it and would also drop `api` from a tab in `api-gateway`, so I have left it.

**The row has no generic fallback.** A tab whose pane says nothing becomes `Shell`. A workspace in the same position keeps the label Herdr gave it, because the directory it was made in says more than `Shell` does. That is a second exclusion beyond the foreground process and you did not ask for it — say if you want `Shell` there instead.

## Issue #80

This covers the workspace half of it: a workspace already named when the plugin starts is never taken. One created after start is named as any other, by the second rule of #69, so a name an orchestrator writes at creation is still replaced — and tabs and panes are unchanged.

## Gate

`make check` green, `golangci-lint` at 0 issues. Five commits, each building and testing green on its own. New tests cover every row of the table above plus the truncation ladder.


6temes (changes requested):
Careful work, and the docs are unusually honest about what the rule as stated
costs. `make check` is green here at `3cbad5b`, golangci-lint at 0 issues.

One defect I think blocks, and it is the one the PR body raises under "the
identity test also stops matching once this has done its job" — I read that as
understating it.

## A row Auto Title wrote is claimed as the user's on the next start

Sequence: the plugin names a row, Herdr keeps that label, the plugin starts
again, and the judging poll sees a label that is neither the directory's
basename nor empty. `Claims.Observe` takes that as a name its owner wrote and
locks it, at the stale label, for good.

The PR body attributes this to `herdr server stop` after a config change. But
AGENTS.md records that Herdr **runs the startup hook again at every start and
live handoff** — which is what `App.superseded` exists for. So this is not a
once-per-config-change cost: a row goes stale and then locks at the first
handoff after it was named, and the lock file fills with labels the plugin
wrote, indistinguishable from real manual names.

A row that is named once and then frozen at a name the work has moved past is
the thing #69 was filed about, so I do not think this can ship as stated.

Reproduced against `3cbad5b`, in `internal/app`:

```go
func TestARowAutoTitleNamedIsNotClaimedAfterARestart(t *testing.T) {
	repo := repoAt(t, "feat/oauth")
	cfg := workspaceConfig()
	cfg.ManualPath = filepath.Join(t.TempDir(), "manual-names.json")

	h := soleTabWorkspace(t, cfg, repo, herdr.PaneInfo{})
	h.polls(2)

	named := filepath.Base(repo) + " › feat/oauth"
	wantRow(t, h, named)

	// Herdr re-runs the startup hook at every start and live handoff, against
	// the same session and the same lock file.
	restarted := startConfigured(t, h.client, cfg)
	restarted.polls(2)

	if restarted.app.manual.Workspaces.Locked(theWorkspace) {
		t.Errorf("the row Auto Title wrote (%q) was claimed as the user's", named)
	}

	// And the row must still follow the work.
	writeHEAD(t, repo, "feat/billing")
	restarted.polls(3)
	// ...assert the last rename is "<repo> › feat/billing"
}
```

Both assertions fail:

```
the row Auto Title wrote ("001 › feat/oauth") was claimed as the user's
after a restart the row ended "001 › feat/oauth", want "001 › feat/billing"
```

### The fix I would suggest

`manualFile` already persists what a poll needs to survive a restart; it just
persists the wrong half for this. If it also carried the label Auto Title last
wrote per workspace, the judging poll could tell its own work from the owner's,
which is the one distinction the basename rule cannot make once the plugin has
written a label of its own. `Claims.Applied` already records exactly that value
in memory.

That keeps the rule as @kryptamine stated it — an unnamed workspace wears its
directory's basename, anything else is the owner's — and adds the case the rule
predates: a label this plugin put there. `TestOnlyTheRowStillWearingItsDirectoryIsTaken`
and `TestALabelThatIsNotTheDirectoryIsTreatedAsTheUsers` should both stay green.
Nothing in the suite pins the current behaviour, so no test has to be rewritten
to allow it.

## What I checked and found sound

- Nil panes: `paneReads.fill` and `Deterministic.collect` both refuse one, so a
  tab with no pane falls out with an empty name rather than a panic.
- `withoutRow` matches with `EqualFold`, which is what `withoutAbove` already
  did, so turning the feature on does not silently change a tab's matching.
- `WorkspaceMaxLength` cannot reach zero: `count` floors at 1 and `Sanitize`
  reads 0 as "no limit", so the guard is in the parser rather than missing.
- `Retain` prunes `pending` and releases a lock whose owner's label moved.
- `Manual.Settled` is still reached only after every workspace was seen, and
  `nameWorkspaces` returning on a cancelled context preserves that.
- Repeated `fill` in one poll is memoised by `changes.Processes` and
  `checkoutMemo`, so naming a row above a claimed tab costs the read the two
  tests about it say it does.
- `FormatTail` measures the sanitized string rather than the raw parts, which is
  the subtle half of the truncation ladder and is tested.

## On the terminal title

You asked whether the row should carry it. I would keep it. The argument in
`title-resolution.md` — that a row outliving every command must not be rewritten
at every prompt — applies to the foreground process, and reading the title
unbound is what separates the two. Dropping it would leave a pane running an
editor with nothing but its directory, which is a worse row than the filename.

chechunhsu:
@6temes — you are right, and "understating" is the word: I had it down as a once-per-config-change cost, and the startup hook running at every handoff makes it the common case. Fixed as you suggested, in `1745bad`: `Claims.Applied` now keeps the label last written per workspace in the lock file (`written_workspaces`), and the judging poll reads a row wearing that label as its own — renamed on, not claimed. It is forgotten with the workspace, and a row the user renames to exactly what the plugin last wrote is not told apart, as a tab renamed to *Desired* is not. A rename that landed after its call got no answer is recorded the same way once a poll sees it on the row. Your test is in as `TestARowAutoTitleNamedIsNotClaimedAfterARestart`, with a state-level pair beside it; the two tests you named stay green.

The terminal title stays in the row, for the reason you give.

internal/state/manual.go
```diff
@@ -142,17 +164,67 @@ func (c *Claims) Observe(s Sighting) bool {
 	previous, known := c.seen[s.ID]
 	ours := c.ours(s)
 
+	// A rename that landed late is as much this plugin's work as one that
+	// answered, and a restart must find it recorded as such.
+	if _, landed := previous.sent[s.Current]; landed && c.lockUnseen {
```
kryptamine, line 169: Small one: this peeks into `previous.sent` directly, and `ours()` two lines up already asked that question.

```go
if c.judgeOnSight && ours && s.Current != "" {
	c.written[s.ID] = s.Current

	m.saveLocked()
}
```

internal/state/manual.go
```diff
@@ -142,17 +164,67 @@ func (c *Claims) Observe(s Sighting) bool {
...
+	}
+
 	delete(previous.sent, s.Current)
 	c.seen[s.ID] = labels{current: s.Current, sent: previous.sent}
 
+	// The poll a workspace is told apart on: the first, or for one that showed
+	// no default then, the first that does. Until then it stays unseen, so that
+	// poll still finds it new.
+	_, pending := c.pending[s.ID]
+	judging := c.lockUnseen && !known && (!m.settled || pending)
```
kryptamine, line 182: My main structural worry. `Observe` grew from 35 lines to 85, and `lockUnseen` is read three times in this function (169, 182, 213).

The two rules are opposites, not variations: for `!known && settled` a tab falls out of the switch and locks at 237, while a workspace returns false at 215, twenty lines apart with the `""`/default check between them. Reading for either, the other is in the way. Ten tests across `internal/app` exist only to pin down when a workspace is judged.

Would you consider reading the flag once, and letting each rule be its own function?

```go
previous, known := c.seen[s.ID]
ours := c.ours(s)
c.record(s, previous, ours)

if c.judgeOnSight {
	return c.claimedOnSight(s, known)
}

return c.claimedOnChange(s, previous, known, ours)
```

`record` takes the bookkeeping both rules share, and needs `ours` for the `written` write. Each `claimed*` stays under fifteen lines, the tab rule returns to its pre-PR shape, and no call site changes. This composes with the `Verdict` suggestion on `app.go:449`: both `claimed*` would return that instead of a bool. `judgeOnSight` is your own word from the commit message ("claim a workspace the user named, on sight"), and names the rule where `lockUnseen` names the mechanism.

internal/app/app.go
```diff
@@ -187,13 +213,126 @@ func (a *App) readAndRename(ctx context.Context, client herdr.Client) error {
...
+		}
+
+		context := contexts[first[info.WorkspaceID]]
+		reads.fill(ctx, client, context)
+
+		ws := state.WorkspaceFrom(info, context)
+		a.manual.Workspaces.Observe(state.WorkspaceSightingFrom(ws, ""))
+	}
+
+	for _, info := range snapshot.Workspaces {
```
kryptamine, line 273: Two loops over the same slice, each with its own `ctx.Err()`, `reads.fill` and context lookup. `workspaceTabs` builds `sole` out of `first` (325-326), so for a sole workspace both passes read the same pane through the same tab.

One loop would do, and not-sole and locked collapse into each other, since both only observe with an empty `Desired`:

```go
tabID, sole := only[info.WorkspaceID]
if !sole || a.manual.Workspaces.Locked(info.WorkspaceID) {
	a.manual.Workspaces.Observe(state.WorkspaceSightingFrom(ws, ""))

	continue
}
```

internal/app/app.go
```diff
@@ -300,6 +444,13 @@ func (a *App) apply(
 		return
 	}
 
+	// Parked without a default, so not yet known to be nobody's: naming it now
+	// would bury the label the judging poll is still waiting to read.
+	if claims.Pending(seen.ID) {
```
kryptamine, line 449: This branch can only fire for a workspace, in the function tabs and panes share, and it takes the mutex again to read what `Observe` computed a line earlier. It is a tri-state expressed as a bool plus a follow-up question.

A `Verdict` return (`VerdictName`, `VerdictClaimed`, `VerdictUnjudged`) folds both: `apply` switches once, `Pending` leaves the public API, the second `Lock()` goes, and `apply` stops knowing workspaces exist. Cost is one line at each of the two existing `Observe` call sites.

internal/resolver/terminal.go
```diff
@@ -28,5 +36,10 @@ func (TerminalTitle) Resolve(pane *state.PaneState) (Parts, bool) {
 		return Parts{}, false
 	}
 
+	// An agent is a field of its own rather than a bound kind, so it stays.
+	if s.unbound && !pane.HasAgent() {
```
kryptamine, line 40: NIT: A field, a second constructor and the split of `activityFrom` to skip one call to `qualify` is a fair trade at three callers against one. The doc comment is what reads oddly: "activityAs is activityFrom under a kind the caller chose" describes the wrapper rather than why the parameter exists.

internal/app/app.go
```diff
@@ -187,13 +213,126 @@ func (a *App) readAndRename(ctx context.Context, client herdr.Client) error {
...
+			state.WorkspaceSightingFrom(workspace, decision.Name),
+			decision,
+		)
+	}
+}
+
+// workspaceTabs maps a workspace to the only tab it holds (sole), which naming
+// skips on when absent, and to its first tab whatever the count (first), which
+// the claim pass reads a directory through.
+func workspaceTabs(tabs []herdr.TabInfo) (sole, first map[string]string) {
```
kryptamine, line 314: Three maps and a `several` set to answer how many tabs a workspace holds. A `count map[string]int` beside `first` answers it directly, with `sole := count[id] == 1` at the call site. That drops a map, the pass that deletes from it, and the comment explaining why there were two.

internal/state/manual.go
```diff
@@ -255,7 +359,12 @@ func (m *Manual) saveLocked() {
...
-		manualFile{Locked: m.Tabs.locked, LockedPanes: m.Panes.locked},
+		manualFile{
+			Locked:            m.Tabs.locked,
+			LockedPanes:       m.Panes.locked,
+			LockedWorkspaces:  m.Workspaces.locked,
+			WrittenWorkspaces: m.Workspaces.written,
+		},
 		"",
 		"  ",
 	)
```
kryptamine, line 370: Config.ManualPath  = "" keeps them in memory only, and DefaultManualPath also returns "" when os.UserConfigDir() fails - something a server-spawned plugin can hit by inheriting the server environment.

In that case, saveLocked returns immediately, written is lost with the process, and the judging poll can claim a row it wrote itself. Your test with cfg.ManualPath = "" reproduces this: Auto Title writes 001 › feat/oauth, then it’s treated as the user’s rename, so wE named "001 › feat/oauth", want "001 › feat/billing".

For a tab, this only loses rename persistence; for a workspace, the row freezes and the feature stops working.

6temes (dismissed):
Clearing my changes-requested: the defect is fixed and I checked it rather than
taking the commit message for it, at `64433ad`.

- The test I posted passes, and `TestOnlyTheRowStillWearingItsDirectoryIsTaken`
  and `TestALabelThatIsNotTheDirectoryIsTreatedAsTheUsers` both stay green.
- Three mutations, all killed: dropping the `written` check from the judging
  poll fails my test and your `TestARowAutoTitleNamedIsNotClaimedAfterARestart`;
  stopping `Applied` from recording the label fails those plus
  `TestARowThisWroteIsItsOwnAfterAReload`; dropping the late-landing record
  fails `TestARowRenameThatLandedLateIsItsOwnAfterAReload`.
- `git diff --stat 3cbad5b 64433ad` is five files — the fix, its docs and its
  tests. Nothing else moved in the rebuild.
- `make check` green, lint 0 issues.

@kryptamine's empty-`ManualPath` case is real and I reproduced it: with
`cfg.ManualPath = ""` my test fails again exactly as he describes, and
`DefaultManualPath` returning `""` when `os.UserConfigDir()` fails is reachable
by a server-spawned plugin. That one still needs an answer.

This approval is scoped to what I raised. The structural review — the
`claimedOnSight`/`claimedOnChange` split, the `Verdict` tri-state, the single
loop in `nameWorkspaces` — is @kryptamine's call, not mine, and I am not
blocking on it.

chechunhsu:
@kryptamine — taking all of it, as commits on top of `64433ad` so @6temes's review stays anchored; I will fold them into the five whenever you want the final shape, and rebase onto 0.8.0 in the same pass (nine files conflict now, all adjacent-line: `app.New` gaining `instance`, the method lists, `notification.show`).

**The empty manual path.** With no file, nothing in memory can tell a row this named from one the user did after a restart, so the row would be named once and then freeze. It is now not named at all when `HERDR_AUTO_TITLE_MANUAL_FILE` names no file, and `LoadConfig` says so with a warning — the same way 0.8.0 degrades when there is no platform directory to claim the session in. Since 0.7.0 the default path is empty only when neither a home nor a platform directory can be found, so the spawned case you describe is narrower than it was; the explicit empty string is the one that remains. Say if you would rather fall back to a temp file instead.

**The structure**, one commit each: `Observe` reads the flag once — `judgeOnSight` — and hands the sighting to `claimedOnSight` or `claimedOnChange` after `record`, the tab rule back in its pre-workspace shape; `Observe` answers a `Verdict`, `Pending` is gone and `apply` switches once; `nameWorkspaces` is one pass with not-sole and locked collapsed, over a `count`; the late-landed record goes through `ours`; the `activityAs` comment says why the parameter exists.

Your `record` sketch changed one behaviour, so it is a commit of its own rather than part of the refactor: recording through `ours` means a row the user renames to exactly the name wanted is now this plugin's across a restart too, as the tab rule already read it while running — before, the file did not agree and the row was claimed at the next start. It asks `ours && known`, not `ours` alone: on the judging poll a label that happens to equal *Desired* was there first, and #69 claims it.

Also in: the Windows CI failure was a test fixture with no drive letter, fixed.

internal/state/manual.go
chechunhsu: Done in 21b34a4 — through `ours`, and a commit of its own since it changes what a restart sees; the refactor in c973615 keeps the `sent` check so it stays behaviour-preserving.

internal/state/manual.go
chechunhsu: Done in c973615: `judgeOnSight` read once, `record` + `claimedOnSight` / `claimedOnChange`, no call site changed.

internal/app/app.go
chechunhsu: Done in cf6ffea: one pass, not-sole and locked collapsed into the empty-`Desired` sighting.

internal/app/app.go
chechunhsu: Done in c973615: `Observe` answers a `Verdict`, `Pending` is gone, `apply` switches once.

internal/resolver/terminal.go
chechunhsu: Done in 9b655b4.

internal/app/app.go
chechunhsu: Done in cf6ffea: `count` beside `first`, `sole`/`several` gone.

internal/state/manual.go
chechunhsu: Done in 6d536ef: with no file the row is left alone and `LoadConfig` warns, rather than named once and frozen — say if you would rather fall back to a temp file.

kryptamine:
@chechunhsu Thanks! Let's leave it as-is for now when `HERDR_AUTO_TITLE_MANUAL_FILE` is empty. Actually, I think we should remove this behavior altogether - keeping the in-memory state only adds complexity without providing much value.

Let's rebase everything now. I'll do one more round of manual testing, and if everything looks good, I'll merge it.

chechunhsu:
@kryptamine — rebased onto `main` at `666f7bc` and folded back into the five commits, each building and testing green on its own; the six follow-ups are in the commit they belong to, their bodies extended to say why. The conflicts were the adjacent-line ones: `app.New` now takes the workspace resolver and the instance, `workspace.rename` sits beside `notification.show` (six methods, `AGENTS.md` and the socket note count them), and the manual file's example text carries both changes.

The empty `HERDR_AUTO_TITLE_MANUAL_FILE` is left as it is: row left alone, warning at load. Removing the in-memory mode altogether reads right to me too — it would take that warning and `namesWorkspaces` with it — but as a change of its own after this lands, if you open an issue for it.

chechunhsu:
@kryptamine — the Windows failure was mine, in `f7e06fb`: `dirBase` took only `/` for a root, so a pane at `C:\` answered a default of `\`, and the fixture I had added appended the volume twice. A root is now the volume plus a separator, unchanged on every other platform. The run needs your approval again.

6temes (approved):
Re-approving after the rebase onto 0.8.0; my approval went with the force-push,
so this is the same scope as before — checked again at `f7e06fb` rather than
assumed to have survived.

The restart defect stays fixed, and the rework strengthened it. The test I
posted still passes, and mutating the recognition in `claimedOnSight` — so the
judging poll no longer knows its own label — now fails five tests rather than
three:

```
--- FAIL: TestARowAutoTitleNamedIsNotClaimedAfterARestart
--- FAIL: TestARowThisWroteIsItsOwnAfterAReload
--- FAIL: TestARowRenameThatLandedLateIsItsOwnAfterAReload
--- FAIL: TestARowRenamedToTheNameWantedIsItsOwnAfterAReload
```

The structural rework reads better than what I reviewed: `claimedOnSight` is
twenty-two lines and its switch says the four cases in order, the `Verdict`
tri-state took `Pending` out of the public API, and `written` is now recorded
through `ours` rather than by peeking into `previous.sent`.

`f7e06fb` is right on the root, and unchanged off Windows:
`filepath.VolumeName` returns `""` on Unix, so `root` stays `/` there.

`make check` green here, lint 0 issues, including the new `internal/instance`
package from 0.8.0.

The only thing I have not seen is CI — the run still needs approval, so treat
this as one macOS gate, not five.

kryptamine (approved):

kryptamine:
Let's get this merged.

PR #75, opened 2026-09-15
fix(app): read the branch from the agent's own directory

### LINKED ISSUE #74: the branch segment never shows for an agent working in a worktree
## What happens

A tab whose agent is working inside a git worktree never names the branch. In a
session of fourteen such panes, every one read alike: the pane reported its
repository root on `main`, while the work sat in `.claude/worktrees/*` on a
feature branch. One tab read

```text
1 · rails › Tokyo dev workspace discussion
```

while the agent in it was on `spec-pin-jobs-landing-contract`.

## Why

The branch source is correct; its input is not. `paneReads.fill` reads one
directory — the pane's foreground process directory — and uses it for the
context, the branch and the transcript lookup alike. A coding agent does not
move the pane's directory: it enters the worktree internally, so that directory
stays at the repository root, `git.Read` answers with the trunk, and the
trunk-suppression rule in `Git.label` deletes the segment. Every rule downstream
behaves exactly as designed.

`title-resolution.md` half-names this case already — *"Agents run in worktrees,
and a worktree's directory is exactly the kind that names nothing"* — but that
sentence is about a worktree the **pane** is in, which is the case that does not
arise for an agent.

## Proposed scope

Read the directory the agent is working in from the Claude Code session
transcript the plugin already opens for the agent's topic (`cwd`, present on
`user`, `assistant`, `system` and `attachment` lines), and resolve the branch
from there when it belongs to the same repository and is not that repository's
trunk. Everything else stays:

- The **context** keeps naming the project from the pane's own directory. A
  worktree directory names nothing, and `withoutRepetition` drops a branch equal
  to its context — a `.claude/worktrees/<branch>` basename *is* the branch, so
  moving the context there would silently delete the segment.
- The transcript supplies a **directory, not a branch**. Its own `gitBranch`
  field reads the literal string `HEAD` in sampled sessions, and trusting it
  would bypass trunk suppression, `BranchMax` shortening, issue-key extraction
  and detached-HEAD handling. The path is the input; `internal/git` stays the
  authority.
- An **override with a fallback**, never a replacement: no repository at the
  agent's directory, a different repository, or a trunk answer all keep the
  pane's own branch, so no pane loses a segment it shows today.
- No new setting. `HERDR_AUTO_TITLE_TRANSCRIPT` already gates the read.

Measured over 496 transcripts on one machine: 317 end at a directory inside a
worktree, 58 end outside any repository, and 15 span more than one repository
root — which is what the same-repository gate is for.

The cost worth naming up front: a branch consumes about 15 of the 50 columns and
truncation cuts the tail, so an agent pane's topic keeps roughly 31 columns where
it kept 44.

Happy to send the change if the scope looks right.

6temes:
## What this changes

`paneReads.fill` read one directory and used it for the context, the branch and
the transcript lookup alike. A coding agent never moves the pane's directory —
it enters the worktree internally — so the branch came back as the trunk and
suppression deleted it. Both checkouts are now read separately, `Git` and
`AgentGit`, and the branch the tab shows is picked in `Git.Resolve` after both
are labelled. The context still comes from the pane's own directory, because a
worktree directory names nothing and `withoutRepetition` would drop a branch
equal to it.

A detached agent worktree still says its hash, and an agent on a trunk still
says nothing — neither has a case in the resolver, because the label already
knows.

Nothing was probed about Herdr's API, so `herdr-socket-api.md` is untouched.

Closes #74

## Since your review

Your five comments, and what each became:

- **The gate re-decided what `label` decides.** `overridesBranch` is gone. The
  pick is your snippet, unchanged.
- **Two plain reads, no third parameter.** Done; `checkout` lost the parameter
  and both shortcuts.
- **The strict path check in `isWorkingDir`.** Gone; the reader reports the
  `cwd` as the transcript spelled it and refuses only an empty value, which
  would erase the directory the session last named.
- **The existence check belongs in `discover`.** Moved. `isDir` is gone from
  `internal/app`, and the refusal now covers every caller.
- **The no-`origin/HEAD` case had no test.** Two now, in both directions, with
  a fixture that removes `refs/remotes/origin/HEAD` — which is why the case was
  unreachable before: every other app fixture writes it.

On the sibling-row asymmetry: left as it is, per your answer.

## One behaviour change worth your attention

Deleting the gate also deletes its no-default arm. In a repository recording no
default branch, `label` suppresses nothing, so an agent standing on `main` now
takes the tab from the pane's own feature branch:

```text
pane on feat/oauth, agent at the root on main, no origin/HEAD
  before  wt › feat/oauth › claude
  after   wt › main › claude
```

It is the same mechanism that finally shows the agent's branch in such a
repository, which is what this source exists for, so the two arrive together.
Pinned by `TestAnAgentOnATrunkNobodyRecordedNamesItAnyway` rather than left
implicit. Treating `main`/`master`/`trunk` as the trunk where none is recorded
would close it, but that is a different defect — the pane's *own* branch reads
`main` there too — and belongs in its own pull request.

One subtlety in `SameRepository` that a future edit could turn into a bug: two
checkouts outside any repository share no common directory and so compare
*equal*. Nothing stops the agent winning there except its label being empty, in
another package. The comment says so, but the day `label` gains a fallback for
an empty checkout, two unrelated panes become one repository.

## How it was checked

`make check` green, `go test -race` re-run uncached. Beyond it:

**A truth table, mechanically.** The old gate and the new pick were
reimplemented as pure functions and every reachable `Checkout` pair enumerated:
289 pairs, 14 raw disagreements. All but two are cases where a branch labels to
nothing and the new code falls back where the old one emitted no branch at all.
The two that remain are the no-default class above, both directions, both
pinned.

**Mutation checks**, because a green suite proves less than a killed mutation:
neutering the pick fails 10 tests; forcing `SameRepository` true fails 4;
dropping the label-emptiness test fails 12; removing the existence refusal
fails 3; accepting an empty `cwd` fails 6.

That last set also caught two of my own tests asserting their fixture rather
than the behaviour they were named for — both survived every mutation,
including one that showed a tab taking the branch of the worktree *above* a
removed one. Fixed in the last commit.

## Unapplied findings

- **`discover`'s existence check catches a path that is gone, and only that.**
  A directory that outlived its own `.git` file still exists, so the walk up
  still answers with the repository above it. Where the main checkout is itself
  on a feature branch, that mislabels the pane. No check there can tell the
  shape from a legitimate subdirectory of a checkout, which this source
  deliberately resolves upward, so the comment and the architecture doc now say
  "gone" rather than claiming more.
- **`commonDirOf` trusts the `commondir` file.** A hand-edited or corrupt one
  spelling `/` makes two repositories compare as one; `.` makes a worktree fail
  to match its own. Git never writes those, but the field was cosmetic before
  this change and is a decision input now.
- **Symlinked spellings still refuse rather than follow**, as before — a silent
  no-op, never a wrong label, and pinned by a test.
- **The feature is inert when the pane's own directory is outside a
  repository**, since the comparison needs the pane's checkout to match. That
  is the shape a directory-of-checkouts pane has, and it is deliberate here.

## Before merging

- [x] No probe taught anything new about Herdr, so nothing was recorded.


internal/app/reads.go
```diff
@@ -151,7 +156,45 @@ func (p *paneReads) checkout(ctx context.Context, dir string) git.Checkout {
...
+func isDir(dir string) bool {
+	info, err := os.Stat(dir)
+
+	return err == nil && info.IsDir()
+}
+
+// overridesBranch reports that the agent's checkout speaks for the branch: it
+// is another view of the same repository, and it has something to name. Why the
+// common directories are not resolved is in docs/architecture/title-resolution.md.
+func overridesBranch(own, agent git.Checkout) bool {
```
kryptamine: This re-decides what `Git.label` in `internal/resolver/git.go` already decides: detached says something, trunk says nothing. It has to, because the choice is made before the label exists.

If the state carries both checkouts and the resolver picks after labelling, this function goes away:

```go
branch := g.label(pane.Git)
if pane.AgentGit.SameRepository(pane.Git) {
      if agent := g.label(pane.AgentGit); agent != "" {
              branch = agent
      }
}
```

internal/app/reads.go
```diff
@@ -103,8 +104,12 @@ func (p *paneReads) fill(ctx context.Context, client herdr.Client, pane *state.P
...
 	pane.Processes = state.ProcessesFrom(processes)
 	pane.Dir = dir
-	pane.Git = p.checkout(ctx, dir)
-	pane.AgentTopic = p.topic(ctx, pane, dir)
+
+	// The transcript is read before the checkout because it says where the
+	// agent is working, and that is where the branch is read from.
+	topic := p.topic(ctx, pane, dir)
+	pane.AgentTopic = topic.Text()
+	pane.Git = p.checkout(ctx, dir, topic.Dir)
```
kryptamine: Suggest two plain reads and no third parameter:

```go
pane.Git = p.checkout(ctx, dir)
pane.AgentGit = p.checkout(ctx, topic.Dir)
```

That way `pane.Git` means one thing again, and the `agentDir == dir` shortcut isn't needed because `checkoutMemo` already collapses equal directories.

internal/claude/transcript.go
```diff
@@ -310,6 +318,13 @@ func (t *transcript) absorb(lines string) {
 	}
 }
 
+// isWorkingDir accepts a directory as the transcript spelled it. The value
+// becomes the directory a checkout is read from, so anything but an absolute,
+// clean path is refused rather than repaired.
+func isWorkingDir(dir string) bool {
```
kryptamine: `git.discover` already rejects relative paths and cleans the rest. The strict clean check was only needed for the `agentDir == dir` comparison, so once that goes, this can go too and the reader can keep `cwd` as written.

internal/app/reads.go
```diff
@@ -151,7 +156,45 @@ func (p *paneReads) checkout(ctx context.Context, dir string) git.Checkout {
...
+	}
+
+	// A detached HEAD always has something to say. A trunk never does, and nor
+	// does a branch in a repository recording no default, which cannot be told
+	// from that repository's trunk — so the pane's own answer stands instead.
+	if agent.Branch == "" {
+		return true
+	}
+
+	return agent.Branch != agent.Default && (agent.Default != "" || own.Branch == "")
```
kryptamine: Without `origin/HEAD` this refuses whenever the pane is on a branch, so the #74 case (pane on `main` at the root, agent in a worktree) still shows `main`. There's no test for it either: replacing the clause with `true` keeps the suite green. The label-based version from the comment above shows the agent's branch here.

internal/app/reads.go
```diff
@@ -151,7 +156,45 @@ func (p *paneReads) checkout(ctx context.Context, dir string) git.Checkout {
...
+		return agent
+	}
+
+	return checkout
+}
+
+// isDir reports whether dir is still there to be read. A worktree the agent
+// left behind is removed while the transcript still names it, and the walk up
+// would answer with the ancestor repository instead.
+func isDir(dir string) bool {
```
kryptamine: A walk up from a removed directory is a `git.discover` problem for any directory. A shell sitting in a deleted directory gets its parent repository's branch the same way today. Could this stat move to the top of `discover`? The existing tests still pass with it.

kryptamine (changes requested):

kryptamine:
On the P2: I'd leave it. withoutAbove already does this with the directory (see the pane-rename / herdr-reviewr example in title-resolution.md), and the branch just follows the same rule.

6temes:
Reworked along your comments — `overridesBranch` is gone and the pick is your snippet, unchanged. The existence check moved into `discover`, `isWorkingDir`'s strict form went with it, and the no-`origin/HEAD` case has tests in both directions now.

Two things worth your eye, both in the body:

Deleting the gate also deletes its no-default arm, so where a repository records no default an agent standing on `main` now takes the tab from the pane's own feature branch. Same mechanism that finally shows the agent's branch there, so the two arrive together — pinned by a test rather than left implicit. Treating `main`/`master`/`trunk` as the trunk where none is recorded would close it, but the pane's *own* branch reads `main` there too, so that felt like its own pull request rather than scope for this one. Happy to fold it in here if you'd rather.

And a subtlety I'd not spotted before: two checkouts outside any repository share no common directory and so compare *equal* in `SameRepository`. Nothing stops the agent winning there except its label being empty, in another package. Harmless today, and the comment says so, but it would become a real bug the day `label` gains a fallback for an empty checkout.

Checking it, a truth table over every reachable `Checkout` pair put the old gate against the new pick: 289 pairs, 14 disagreements, and all but two are cases where a branch labels to nothing and the new code falls back where the old one showed no branch at all. The two left are the no-default class above.

That pass also caught two of my own tests asserting their fixture instead of the behaviour they were named for — both survived every mutation of the code they covered, including one that showed a tab taking the branch of the worktree *above* a removed one. Fixed in the last commit; the same failure mode you caught in the first round, so thanks for the nudge on that.

kryptamine (dismissed):

kryptamine (approved):

kryptamine:
The no-default behavior makes sense, and I agree the main/master/trunk fallback belongs in a separate PR.
The `SameRepository` edge case looks fine as-is, and good catch on those tests.
I fixed the flaky test in a separate commit and merged the PR.

PR #77, opened 2026-09-15
fix: stop a rename that lands after its call failed locking the tab

### LINKED ISSUE #76: A rename that lands after its call timed out locks the tab on a stale name
### What happened

Tabs kept a stale position after a tab to their left closed. A tab sitting
fifth read `6 · …`, two tabs read `6 · …` at once, and a tab at position 1
read `3 · …`. `manual-names.json` held twelve locked tabs, and every label in
it had the `N · title` shape Auto Title gives a tab itself — nobody had
renamed any of them by hand. A locked tab is never renamed again, so its
number froze for good.

### What you expected instead

A rename Auto Title made is never taken for the user's, however late Herdr
applies it, so a tab's position keeps following it.

### How to reproduce it

It needs the Herdr server to stall for longer than a poll's five seconds,
which cannot be arranged on demand; this is the sequence it went through:

1. A tab reads `3 · api`, and a tab to its left closes. Auto Title sends
   `tab.rename` with `2 · api`.
2. The server is stalled. Herdr has read the request but carries it out only
   seconds later, after the call has hit the poll's deadline. `apply` sees an
   error and records nothing.
3. Meanwhile the name wanted moves on again — another position change, or a
   poll whose reads are cut short.
4. The late rename lands. The tab now carries a label that is neither the one
   last seen nor the one wanted, so `Claims.Observe` reads it as the user's
   and locks the tab.

The fix comes with a regression test that models this with a stub rename
which lands and still reports an error.

### Where it comes from

Herdr's server log records when it receives a request and when it applies it.
Of 977 `tab.rename` requests in the log, 12 were applied more than a second
after they were received — 3 to 18 seconds — and 10 of those 12 are the last
rename of a locked tab, the lock taken on exactly that label.

`App.apply` calls `Claims.Applied` only when `tab.rename` answers. A call that
gives up has not undone its request: Herdr carries out a request it has read
even after the caller has hung up. The protection then compares a label that
landed late against a *Seen* and a *Desired* that have both moved on.

The stalls themselves are Herdr's: for their whole length the server logged
next to nothing, and one of them ended the moment a terminal client that had
stopped reading disconnected. Auto Title cannot prevent them, but it should
survive them.

### Auto Title version

0.6.0, and a linked checkout of `main`; the code path is the same at 2700769.

### Herdr version

herdr 0.9.0-preview.2026-09-08-62431dbd033b

### Operating system

Windows 11, where it was seen. Nothing in the path is platform-specific: any
server stall longer than a poll does the same on macOS and Linux.

### Configuration

```shell
HERDR_AUTO_TITLE_AGENT_NAME=false
```

### What you tried

Clearing each locked tab's name hands it back, and the tabs were renamed to
their positions on the next poll. That fixes the tabs already locked, not the
cause. The fix keeps the label of every failed rename per tab and per pane,
and a tab found wearing one is Auto Title's, like one wearing *Desired*. The
label is forgotten once it lands or the tab closes. Recording it as *Seen* up
front instead would be wrong the other way: a rename that never lands would
make the old label look moved, and lock that.

recih:
## What this changes

A rename that lands after its call failed no longer locks the tab. A poll
gives up on a call after five seconds, but Herdr carries out a request it has
read even once the caller has hung up: in the server log, a stalled server
applied `tab.rename` 3 to 18 seconds after receiving it, and 10 of the 12
tabs locked in that session had such a late rename as their last one, locked
on exactly that label. By the time it lands the name wanted has usually moved
on — a tab to the left closed and every position slid — so the protection
sees a label that is neither *Seen* nor *Desired* and takes it for the user's.
`App.apply` now hands the label of every failed rename to `Claims.Sent`, and
`Claims.Observe` treats a label found there as Auto Title's own, like one
equal to *Desired*, so the tab is renamed on; the label is dropped once it
lands or the tab closes. Recording the label as *Seen* up front would fail
the other way: a rename that never lands would make the old label look moved,
and lock that. The finding about Herdr is recorded in `AGENTS.md` and
`docs/architecture/herdr-socket-api.md`, the rule in
`docs/architecture/manual-rename-protection.md`. Tabs and panes share the path.

Closes #76

## How it was checked

- `gofmt`, `go vet`, `golangci-lint` and the whole suite under `go test -race`
  on Windows, plus `go vet` on the Go 1.24.0 floor and for linux and darwin.
- Two regression tests, each seen failing before the fix: one in `state` for
  a label sent and then observed, one in `app` driving the whole poll through
  a stub rename that lands and still reports an error, while the tab's context
  moves on.
- The cause was read from Herdr's server log, pairing each `tab.rename`
  request with the rename it caused.
- Live: the twelve locked tabs were handed back by clearing their names, and
  this build now runs in the session the bug was found in.

## Before merging

      Windows ships no `make`).


internal/state/manual.go
```diff
@@ -32,10 +32,18 @@ type Claims struct {
 	// locked is the label a thing carried when the user claimed it. The label,
 	// not the id, is what makes a reloaded lock safe: Herdr reuses ids.
 	locked map[string]string
+	// sent holds the labels of renames whose call failed. Herdr may apply one
+	// anyway, seconds later, and it is Auto Title's label all the same.
+	sent map[string]map[string]struct{}
```
kryptamine: Claims now has three maps keyed by the same id, and `Retain` has a third prune loop to match. apply always calls `Observe` before a rename, so an id in sent is always in seen too. Could sent be a field of the seen entry instead?
```go
type label struct {
      seen string
      sent map[string]struct{}
}
```
Then closing a tab drops both in one delete, and the extra loop at [L227](https://github.com/kryptamine/herdr-auto-title/pull/77/changes#diff-5138bb535606e9f196d629db325d586626e9725083177766ee421540df401747R227) goes away.

internal/app/app.go
```diff
@@ -309,6 +309,9 @@ func (a *App) apply(
 			return
 		}
 
+		// A call that timed out may still have reached Herdr, which applies it
+		// whenever it next gets to it: docs/architecture/manual-rename-protection.md.
+		claims.Sent(seen.ID, decision.Name)
```
kryptamine:
```go
switch herdr.ErrorCode(err) {
case kind.gone:
      a.log.Debug(...)
      return
case "":
      claims.Sent(seen.ID, decision.Name)
}
```

internal/herdr/herdrtest/stub.go
```diff
@@ -139,6 +140,15 @@ func (s *Client) SetRenameError(err error) {
 	s.renameErr = err
 }
 
+// SetRenameLost makes every subsequent tab rename land and still fail, as one
+// does when Herdr applies it after the caller stopped waiting for the answer.
+func (s *Client) SetRenameLost(err error) {
```
kryptamine: NIT: `SetRenameError` and `SetRenameLost` are now two separate switches, and `renameErr` wins when both are set. Would one `SetRenameError(err, landed bool)` read better? Also, the doc comment could mention that it only affects tab renames. Not blocking.

internal/state/manual.go
```diff
@@ -131,6 +139,11 @@ func (c *Claims) Observe(s Sighting) bool {
 	previous, known := c.seen[s.ID]
 	c.seen[s.ID] = s.Current
 
+	if _, sent := c.sent[s.ID][s.Current]; sent {
```
kryptamine: The sent check sits outside the switch, so the rule for whose label this is now lives in two places. If you go with the label struct, a small helper such as c.ours(s) could check Seen and Sent together.

kryptamine (changes requested):

kryptamine:
Thanks again for the contribution! Really appreciate you digging into this.

Would you be okay with me adding you as a reviewer for PRs from other contributors?

internal/state/manual.go
recih: Done in df6024e: `seen` is now `map[string]labels`, holding the `current` label and the `sent` ones, so a closed tab drops both in one delete and the third loop in `Retain` is gone.

internal/app/app.go
recih: Done in 44eeb08. This also narrows the behaviour: an `APIError` means Herdr answered and refused the rename, so only a call with no answer is kept in `Sent` now. Added `TestARenameHerdrRefusedIsNotTakenForItsOwn` for that (it fails without the change) and updated the two architecture docs to match.

internal/state/manual.go
recih: Done in df6024e: `c.ours(s)` checks *Desired* and the sent labels together, so the switch opens with a single `case ours:`.

internal/herdr/herdrtest/stub.go
recih: Done in 4a1d45b: one `SetRenameError(err, landed bool)`, and the doc comment now says pane renames are unaffected.

recih:
Thanks for the review! All four points are addressed — see the replies in each
thread — and CI is green. Could you take another look when you have a moment?

And thank you for the offer — I'd be happy to help review PRs from other
contributors. I can't promise a fast turnaround, though; I'll get to them as
my time allows.

kryptamine (approved):

PR #85, opened 2026-09-16
fix(resolver): suppress an unrecorded trunk

6temes:
## What this changes

A repository with no remote records no default branch, so `label` had nothing to
compare against and every tab in such a repository read `main` as work worth the
tab's width. Nothing else in the repository says which branch the others were cut
from, so there alone a name only a trunk carries — `main`, `master`, `trunk` — is
taken to be one. Compared exactly, so `Main` is still a branch.

Measured on my own machine: two repositories in daily use have no remote at all,
and five tabs across them carried a `main` segment that said nothing.

```text
1 · infra › main › Read mainframe fans      →  1 · infra › Read mainframe fans
2 · infra › main                            →  2 · infra
1 · main › Next steps and LFG opportunities  →  1 · Next steps and LFG opportunities
```

## This closes the consequence disclosed in #75

That pull request deleted a gate whose trailing clause refused a branch in a
repository recording no default. Its own body named the cost: an agent standing
on `main` there took the tab from the pane's own feature branch, pinned by
`TestAnAgentOnATrunkNobodyRecordedNamesItAnyway` rather than left implicit.

This is the other half. With a conventional name read as a trunk, the agent's
`main` labels to nothing and the pane keeps its branch, so that test flips to
`TestAnAgentOnATrunkNobodyRecordedLeavesThePanesBranch` and asserts the
behaviour the gate used to give. The equivalence between labelling first and the
deleted gate now holds with no limit, and the architecture doc says so.

## The one judgement in it

This reads a name where the rest of the source reads a ref, which the trunk rule
deliberately did not do — `docs/architecture/title-resolution.md` still says the
trunk "comes from the repository itself, rather than from a list of names", and
that stays true wherever a default is recorded. The list applies only where
there is no recorded default to compare against, which is the one case with no
ref to read.

It also deletes `TestARepositoryWithNoRecordedTrunkAlwaysShowsItsBranch`, which
asserted the opposite. It carried no comment saying why, so if that behaviour was
deliberate rather than the absence of this rule, this is the wrong change and I
would rather hear it than have it merged.

A repository that *has* a remote but has never fetched `origin/HEAD` gets the
same treatment. That is the same silence a fresh clone would get once it fetches,
so it seemed the better default, but it is a choice rather than a necessity.

## How it was checked

Mutation: dropping `isConventionalTrunk` from the suppression fails
`TestARepositoryRecordingNoDefaultStillHasATrunk` on all three names at the
resolver level, and `TestAnAgentOnATrunkNobodyRecordedLeavesThePanesBranch` end
to end — so the rule is pinned at both levels rather than only where it is
written.

## Before merging


kryptamine (approved):

kryptamine:
@6temes LGTM! Feel free to merge when you're ready.

PR #86, opened 2026-09-16
feat(claude): search every configured Claude config home

6temes:
## What this changes

The transcript root came from the plugin's own environment, so a session filed
under a second Claude configuration home was invisible. Herdr's server passes
`CLAUDE_CONFIG_DIR` through to the plugin, and that is one directory — but a user
can have more than one home, and `direnv` or a wrapper can pick a different one
per directory, which the plugin never sees.

`HERDR_AUTO_TITLE_CLAUDE_DIRS` names the extra homes, path-list separated (`:`,
or `;` on Windows, via `filepath.SplitList` so a Windows drive colon survives).
`locate` tries each home in order and `locateUnder` does one home's own two-step
lookup, so a session found in the first home costs exactly what it cost before
this existed.

Measured on my own setup: five panes read no topic at all, their tabs showing
their tabs naming the directory and nothing about the work, because those sessions live under a
second home that a `direnv` rule selects. With the setting they read normally.

## Two decisions worth naming

**The homes are named, not discovered.** Scanning for anything shaped like a
configuration home would read directories the user never pointed the plugin at,
and this source already reads what the user said to their agent — widening what
it opens on its own initiative seemed the wrong direction. It is also why a
refused entry warns rather than failing silently.

**No `Config` field.** The value reaches `claude.NewReader()` through the
environment exactly as `CLAUDE_CONFIG_DIR` does, and it is read in
`internal/claude` where it is used. A field on `Config` would have had no
reader, which the type rule forbids — so the constant is declared in the package
that reads it and aliased in `internal/app` beside the others, for the warning.

A refused entry is skipped and the rest of the list is still read, so one typo
does not cost the homes after it.

## A doc bug found in passing

`docs/architecture/configuration.md` said "one of nine `HERDR_AUTO_TITLE_*`
variables". Ten were declared before this change. It now says eleven and names
the one that lives outside `internal/app/config.go`.

## How it was checked

Mutation: reading only the first configured home (`r.roots[:1]`) fails four
tests — a session found only in an extra home, a refused home still letting the
usable one through, an extra home's transcript continuing to be read across
polls, and a session that moved homes being found again.

The cost when a session is in none of the configured homes is one `projects/*`
glob per home per retry, which is the same shape the single-home lookup already
had.

## Before merging


6temes:
Superseded by #89 — same commits, same tree, but on a branch of this repository rather than my fork so it can be rebased in place. Nothing here was reviewed yet, so nothing is lost with the number.

PR #87, opened 2026-09-16
fix(resolver): follow the agent from a pane holding no repository

6temes:
## What this changes

A pane whose own directory is a parent holding several checkouts shares no
common directory with its agent, so `SameRepository` refused before anything
else was looked at and the tab named no branch at all. #75 recorded that as this
source's boundary rather than a case to widen. This widens it, deliberately.

Such a pane has nothing of its own to lose, so containment decides instead: the
agent's answer is taken when its directory lies under the pane's.

```text
pane at ~/projects, holding dashboard/ and billing/
agent in dashboard/.claude/worktrees/node on chore/node-24.21.0

  before  1 · projects › claude
  after   1 · projects › NODE-24 › claude
```

It is the shape one-pane-per-project produces most often, at least here: my own
sidebar has a pane parked in a directory of checkouts, and it gained no branch
at all before this.

## The three things that decide it

**Belonging becomes two tests, not one.** `belongsToPane` keeps
`SameRepository` for a pane that has a checkout, and falls back to containment
only for a pane that has none. The pick still happens after both labels exist,
so a trunk and a detached HEAD need no case here — unchanged from #75.

**Containment is offered only to a pane with no checkout.** A pane that has one
must not gain a nested clone's branch this way, which is real wherever a vendored
dependency carries its own `.git`. That condition is load-bearing rather than
defensive: dropping it fails
`TestAPaneInARepositoryRefusesAnAgentInOneNestedUnderIt`, where the pane picks
up `fix/token` from `vendor/other`.

**It is measured, not spelled.** `filepath.Rel` plus a refusal of any `..`,
because `code-review` begins with `code` and is not inside it. Replacing it with
a cleaned string prefix fails `TestADirectoryNamedLikeThePanesIsNotInsideIt`.

## Where this needs `AgentDir`

The resolver had the agent's *checkout* but not the directory it was read from,
and containment is a question about directories. `PaneState.AgentDir` carries it,
set beside `AgentGit` in the same read and used only here.

## How it was checked

Three mutations, each killed:

| mutation | fails |
|---|---|
| containment arm removed | `TestAPaneHoldingSeveralRepositoriesFollowsItsAgentsWorktree` |
| the no-checkout condition dropped | `TestAPaneInARepositoryRefusesAnAgentInOneNestedUnderIt` |
| `within` as a cleaned string prefix | `TestADirectoryNamedLikeThePanesIsNotInsideIt` |

`TestAPaneOutsideARepositoryFollowsNoAgent` still passes unchanged — its agent
sits in a sibling tree, so containment refuses it for a sharper reason than
before. Its comment now says which.

## Interaction with the other open pull requests

Independent of #86 (different package). Against #85 there is one overlap worth
naming rather than discovering at merge: a pane outside a repository whose agent
is in a repository recording no default currently takes the agent's `main`
through containment, and #85 would suppress that name. Whichever lands second,
the behaviour is the one its own tests assert; I have kept a test for that pair
out of this branch so neither PR has to be rebased for the other.

## Before merging


6temes:
Superseded by #90 — same commits, same tree, but on a branch of this repository rather than my fork so it can be rebased in place. Nothing here was reviewed yet, so nothing is lost with the number.

PR #89, opened 2026-09-17
feat(claude): search every configured Claude config home

6temes:
## What this changes

The transcript root came from the plugin's own environment, so a session filed
under a second Claude configuration home was invisible. Herdr's server passes
`CLAUDE_CONFIG_DIR` through to the plugin, and that is one directory — but a user
can have more than one home, and `direnv` or a wrapper can pick a different one
per directory, which the plugin never sees.

`HERDR_AUTO_TITLE_CLAUDE_DIRS` names the extra homes, path-list separated (`:`,
or `;` on Windows, via `filepath.SplitList` so a Windows drive colon survives).
`locate` tries each home in order and `locateUnder` does one home's own two-step
lookup, so a session found in the first home costs exactly what it cost before
this existed.

Measured on my own setup: five panes read no topic at all, their tabs naming the
directory and nothing about the work, because those sessions live under a second
home that a `direnv` rule selects. With the setting they read normally.

## Two decisions worth naming

**The homes are named, not discovered.** Scanning for anything shaped like a
configuration home would read directories the user never pointed the plugin at,
and this source already reads what the user said to their agent — widening what
it opens on its own initiative seemed the wrong direction.

**A home is judged by the filesystem, not by its spelling.** A trailing
separator is cleaned rather than refused: tab completion and `$PWD` both produce
one and it names the same directory. An entry that is not a directory is
reported and still searched, because the user can create it a minute after the
plugin started and a miss costs one `projects/*` glob per retry.

## How it was checked

`make check` green, `go test -race` re-run uncached. Each of the two commits is
green on its own.

Mutation: reading only the first configured home (`r.roots[:1]`) fails four
tests — a session found only in an extra home, an unusable home still letting the
usable one through, an extra home's transcript continuing to be read across
polls, and a session that moved homes being found again. Dropping
`filepath.Clean`, dropping an unreadable home from the list, and cutting the
`cfg.ClaudeDirs...` wiring each fail a test of their own.

## Before merging

---

Reopened from a branch on this repository rather than my fork (same commits,
same tree) so the branch can be rebased in place. Replaces #86, which carried
no review or discussion.


internal/claude/transcript.go
```diff
@@ -321,6 +367,13 @@ func (t *transcript) absorb(lines string) {
 	}
 }
 
+// isConfigHome accepts a configuration home as the setting spelled it. The
+// value becomes a directory that is searched, so anything but an absolute,
+// clean path is refused rather than repaired.
+func isConfigHome(dir string) bool {
```
kryptamine, line 373: I think this check may be rejecting homes that would actually work. `filepath.Clean`
strips a trailing slash, so `/home/you/.claude/` lands in `refused` even though the
directory exists and holds the sessions. Tab completion and `$PWD` both produce that
slash, so it's a fairly likely way for someone to write the value.

The reverse case is the one I'd worry about more: `/home/you/.clade` is absolute and
clean, so a typo passes the gate with no warning. `TestAConfigHomeThatCannotBeUsedIsSkipped`
lists `filepath.Join(t.TempDir(), "never-created")` under `unusable`, but it goes into
`dirs` rather than `refused`. So the warning fires on input that works and stays quiet
on input that doesn't.

A relative path already costs nothing, since `Stat` and `Glob` miss it and the home is
skipped. Would it be simpler to drop the spelling check and warn on `os.Stat` + `IsDir`
instead, keeping the home in the list either way? The directory can always appear later.

internal/claude/transcript.go
```diff
@@ -32,6 +32,11 @@ const maxOpening = 200
 // so the search is repeated — just not twice a second for the pane's whole life.
 const locateRetry = 10 * time.Second
 
+// EnvExtraRoots names configuration directories to read besides the one
+// CLAUDE_CONFIG_DIR gives, separated by the platform's path-list separator. A
+// home the user's shell picks per directory cannot be discovered, so it is named.
+const EnvExtraRoots = "HERDR_AUTO_TITLE_CLAUDE_DIRS"
```
kryptamine, line 38: This is the only `HERDR_AUTO_TITLE_*` variable read outside `internal/app/config.go`.
Could it live there like the rest?

Two things that would go away:

1. The double parse. `LoadConfig` keeps only `refused`, `NewReader` keeps only `extra`.
   The two-value return on `ExtraRoots` exists because the parse happens twice.
2. A hidden ordering rule. `readConfigFile` loads `config.env` into the environment and
   `NewReader` reads it later, which only works because `main` calls `LoadConfig` first.
   Nothing in `internal/claude` says so.

It would also let the tests stop juggling the variable: `newProject` clears it defensively,
and `TestLoadConfigAcceptsConfigHomesItCanUse` can only check that no warning appeared,
since `LoadConfig` throws the value away.

internal/claude/transcript.go
6temes: You're right on both halves, and the second one is the worse bug.
`TestAConfigHomeThatCannotBeUsedIsSkipped` did list a never-created directory
under `unusable` while expecting it in `dirs` — the test agreed with the code
that a typo is fine.

Done as you suggested, with one addition: `filepath.Clean` now normalises the
entry rather than judging it, so a trailing separator is read instead of merely
tolerated, and `os.Stat` + `IsDir` decides the warning. The home stays in the
list either way, since it can be created after the plugin starts and a miss
costs one glob per retry.

The spelling check is gone entirely, so `internal/claude` no longer refuses
anything — it searches what it is handed and misses harmlessly. There's a test
for the trailing separator specifically, and dropping the `Clean` call fails it.
Windows caught a genuine mistake in the other new test while I was at it: the
warning formats the path with `%q`, which escapes backslashes, so matching the
raw path only ever worked on POSIX.

internal/claude/transcript.go
6temes: Moved, and the `Config` objection turned out to dissolve rather than hold: once
the homes are passed to `NewReader(extra ...string)` instead of read from the
environment, `Config.ClaudeDirs` has a reader in `newPaneReader`, so the type
rule is satisfied. My "no field" reasoning assumed a field nothing would read,
which is only true if the value keeps travelling by environment.

Both things you named are gone: `ExtraRoots` and its two-value return no longer
exist, and the parse happens once in `LoadConfig`. The tests stop juggling the
variable too — `newProject` no longer clears it, and the app-level tests set
`cfg.ClaudeDirs` directly.

One thing I'd rather not fold in here: the ordering rule isn't specific to this
setting. `roots()` still reads `CLAUDE_CONFIG_DIR` from the environment inside
`internal/claude`, so "`LoadConfig` must have run" is still an undocumented
precondition — just not for `CLAUDE_DIRS` any more. Fixing that properly means
routing Claude Code's own variable through `Config` as well, which is a
different change with its own reasoning. Happy to open an issue for it if you
agree that's the right shape.

Worth flagging that this refactor opened a coverage hole I had to close: once
the environment stopped carrying the value implicitly, nothing proved the homes
reached the reader at all. Cutting the `cfg.ClaudeDirs...` wiring passed every
test until I added `TestASessionInAnExtraConfigHomeIsNamed`.

kryptamine (approved):

recih (approved):

internal/claude/transcript.go
6temes: Filed as #101, with the shape sketched out — read it in `LoadConfig` and pass the base home through `Config` too, so `internal/claude` reads no environment at all. Not urgent: there's no user-visible symptom today, only the hidden ordering rule and the inconsistency of two sibling settings arriving by two routes.

kryptamine: Thanks!

PR #92, opened 2026-09-17
feat: restart the plugin without a server stop

### LINKED ISSUE #91: Restarting Auto Title needs a server stop
### The problem

Getting a fresh Auto Title costs the whole Herdr session. Herdr runs a
plugin's startup hook at server start and live handoff and at nothing else —
not on `plugin install`, `link` or `enable`, and not on a configuration
reload. So each of these needs `herdr server stop` and a new `herdr`:

- the first start after installing or linking the plugin,
- picking up a new version,
- picking up a change to `config.env`,
- recovering an instance that has stopped naming tabs.

Stopping the server ends every pane in it, which is a lot to ask for a tab
namer.

### What it should do instead

A plugin action, `herdr.auto-title.restart`, that starts a fresh instance in
place of the running one and says how that went:

```sh
herdr plugin action invoke herdr.auto-title.restart
```

or on a key, through Herdr's `config.toml`:

```toml
[[keys.command]]
key = "prefix+R"
type = "plugin_action"
command = "herdr.auto-title.restart"
```

The tab bar is unchanged by the restart itself; the tabs are renamed by the
new instance from its first poll, and a notification reads "Auto Title
restarted" or says why it did not.

Herdr keeps no handle on a plugin process, so the old instance has to be told
to leave by Auto Title itself:

- A running instance claims its session in a file under Auto Title's own
  configuration directory, named after the socket path so two sessions never
  displace each other, and reads the claim before every poll. A newer claim
  ends the run; newest wins.
- The new instance waits for the old pid to exit before it loads the
  manual-name locks and polls, and never kills it.
- The action spawns the new instance detached, with its stdio on the null
  device: Herdr reads an action's output to EOF, so a child holding the pipes
  would keep the action "running" and hold one of Herdr's 32 plugin command
  slots. It waits up to 15 s for the new instance to be polling and the old
  one to be gone, then reports through `notification.show` and the action log.

Facts this rests on, read from Herdr 0.9.0's source and checked live on
Windows with a throwaway plugin:

- Herdr rejects a duplicate action id even when the entries' `platforms` do
  not overlap, so one entry has to serve every platform.
- On Windows, Herdr 0.9.0 resolves `./herdr-auto-title` to
  `herdr-auto-title.exe`, so that one entry works there; older Herdr on
  Windows cannot run the action.
- `notification.show` takes `title` and `body` and answers with `shown` and a
  `reason`; not being shown (no client, rate limit) is not an error.

Two limits remain: the first upgrade from a version without the claim still
needs `herdr server stop`, since the running instance does not know to look,
and the action needs Herdr 0.9.0 or newer on Windows.

### What you tried

`herdr server stop` and a fresh `herdr`, every time. `herdr plugin disable`
and `enable` only flip a registry flag and start nothing.

recih:
## What this changes

Auto Title can now be restarted without `herdr server stop`. Herdr starts a
plugin at server start and at nothing else, so a first start after install, a
new version, a `config.env` change and a hung instance each cost the user
their session. A running instance now claims its session in
`instances/<hash of the socket path>.json` under Auto Title's configuration
directory and reads the claim before every poll; a newer claim ends the run.
The new instance waits up to ten seconds for the old pid to leave before it
loads the manual-name locks and polls, never kills it, and marks a separate
`.ready.json` after its first snapshot — separate so marking ready can never
write over a claim taken meanwhile. The `restart` action spawns that instance
detached with its stdio on the null device (Herdr reads an action's output to
EOF, so an inheriting child would hold the action "running" and one of its 32
command slots), waits up to 15 s for it to be polling and the old one gone,
and reports through `notification.show`, Auto Title's fifth socket method,
and the action log. The Herdr facts behind it — startup hooks only at server
start, duplicate action ids rejected across platforms, `./herdr-auto-title`
resolving to the `.exe` on Windows from Herdr 0.9.0, the shape of
`notification.show` — are recorded in `AGENTS.md` and
`docs/architecture/herdr-socket-api.md`; the takeover is in
`docs/architecture/poll-loop.md`, and the README has the action and a
keybinding.

It is one commit rather than two: the claim's tests start real processes
through the same helper the restart uses, so the split did not divide
cleanly.

Closes #91

## How it was checked

- `gofmt`, `go vet`, `golangci-lint` and the whole suite under `go test -race`
  on Windows, plus `go vet` for linux and darwin and a build on the Go 1.24.0
  floor.
- `internal/instance` is tested with real processes: the test binary re-runs
  itself as an instance that claims, waits, marks ready and leaves when
  displaced, one that never leaves, and one that exits at once — covering a
  handover, a first start, an instance dying before it polls, an old instance
  that stays, stale and missing claims, and release leaving a newer claim
  alone. The poll loop's side is tested through a fake claim in `app`.
- A throwaway plugin first confirmed on Herdr 0.9.0 for Windows that
  `./probe` runs `probe.exe` and that `notification.show` answers as read
  from the source.
- Live, on Windows in a real session: the action returned in about half a
  second without being held by the child, a second restart displaced the
  first instance within a second, and tabs were named from the first poll.
  A Herdr server restart later started the linked build, which displaced the
  instance the action had started, leaving one running.

## Before merging

      Windows ships no `make`).


kryptamine:
Thanks a lot @recih  I;ll try to get to this PR tomorrow, unless @6temes  gets to it first :)

internal/instance/instance.go
```diff
@@ -0,0 +1,205 @@
...
+	if c.path == "" || c.names(marker(c.path)) {
+		return
+	}
+
+	_ = write(marker(c.path), record{PID: c.pid})
+}
+
+// Release removes the claim and the marker if they are still this process's
+// own, so leaving because it was displaced does not take the successor's along.
+func (c *Claim) Release() {
```
kryptamine: `Release` checks that the claim is still its own with `names` and then calls `os.Remove`. A successor can write its claim between the two, and the old instance deletes it. This happens in practice. A new server starts instance B, which takes the claim. Instance A leaves through `superseded`, which `poll` checks before `Taken`, and its `Release` can remove B's claim.

`Taken()` (`instance.go:96`) repairs that by rewriting the claim when the file is missing. So a predicate writes to disk to fix a race that lives in another method, and poll-loop.md documents the pairing ("an instance leaving removed it a moment after losing it").

Removing the claim on exit isn't needed at all. `holder()` already checks `alive(pid)`, so a file naming a dead pid displaces nobody. That is how a crashed instance is handled today. Without `Release`, the race, the write in `Taken`, `exists` and `names` all disappear, and `Taken` becomes a `read` and a comparison. The cost is pid reuse after a clean exit.

internal/instance/restart.go
```diff
@@ -0,0 +1,112 @@
...
+			)
+		case state := <-exited:
+			return 0, fmt.Errorf(
+				"the new instance %d exited before it read the session: %v",
+				proc.Pid,
+				state,
+			)
+		case <-deadline.C:
+			return 0, stalled(path, proc.Pid, old, timeout)
+		case <-ticker.C:
```
kryptamine: This checks `holder(path) == pid && ready(path) && (old == 0 || !alive(old))` on every tick, and `stalled()` works the same three conditions out again to report which one failed. `ready()` also calls `holder()` a second time. One `pending(path, pid, old) error` that returns the first unmet condition, or nil, serves both: the tick becomes `if pending(...) == nil` and the deadline becomes `return pending(...)`, so `stalled` goes away.

internal/instance/instance.go
```diff
@@ -0,0 +1,205 @@
...
+		return true
+	}
+
+	return await(ctx, c.displaced, timeout)
+}
+
+// Taken reports whether another process has claimed the session since. A claim
+// that cannot be read decides nothing: it may be half written, and the next
+// look will read it whole. A claim that is gone is nobody's, and taken again.
+func (c *Claim) Taken() bool {
```
kryptamine, line 89: This predicate writes to disk: when the file is missing it writes the claim again. It is there to repair the race in `Release`, and poll-loop.md documents the pairing ("an instance leaving removed it a moment after losing it"). With `Release` gone this becomes a `read` and a comparison, with no side effect.

internal/instance/instance.go
```diff
@@ -0,0 +1,205 @@
...
+// ready reports whether the instance holding the claim has read the session.
+func ready(path string) bool {
+	pid := holder(path)
+	rec, ok := read(marker(path))
+
+	return pid != 0 && ok && rec.PID == pid
+}
+
+// await reports whether the process left within timeout.
+func await(ctx context.Context, pid int, timeout time.Duration) bool {
```
kryptamine: This loop and the one in `restart` are the same shape: a timer, a ticker, `ctx.Done` and a condition. A single `waitUntil(ctx, timeout, cond)` can back both, with `exited` read without blocking inside the restart's condition.

internal/instance/restart.go
```diff
@@ -0,0 +1,112 @@
...
+	"context"
+	"errors"
+	"fmt"
+	"os"
+	"time"
+)
+
+// restartTimeout bounds a restart: the displaced instance gets ten seconds to
+// leave and the new one a first poll after that.
+const restartTimeout = 15 * time.Second
```
kryptamine: The comment says "ten seconds to leave and a first poll after that", which is `app.LeaveTimeout + app.PollTimeout` written out by hand in another package. Change `PollTimeout` and a restart starts reporting failure for an instance that simply hasn't finished yet. Pass the budget in from `main` (`instance.Restart(ctx, path, exe, app.LeaveTimeout+app.PollTimeout)`) or keep one named constant in `app`. The private `restart(..., timeout)` that exists for tests then isn't needed either.

internal/app/app.go
```diff
@@ -113,13 +129,18 @@ func (a *App) Run(ctx context.Context, client herdr.Client) {
...
 // No failure is fatal — Herdr's socket can lag the process it just launched —
-// but another server on the socket ends the run: docs/architecture/poll-loop.md.
+// but a successor ends the run, on the socket or on the claim: docs/architecture/poll-loop.md.
 func (a *App) poll(ctx context.Context, client herdr.Client) bool {
 	if a.superseded(client) {
 		a.log.Info("another server holds the socket and starts an auto title of its own, leaving")
 		return false
 	}
 
+	if a.instance.Taken() {
```
kryptamine: This is a second reason to leave right after `superseded(client)`, each with its own log line and `return false`. When the claim works it already covers a server change, because the new server's instance claims the same socket hash, and the order of the two checks is what opens the race in `Release`. At minimum, fold them into one `a.successor(client) (reason string)` so `poll` has a single exit. Better, decide whether `superseded` is still needed with a claim in place. If it only matters when there is no claim, it belongs with the no-op claim, and the code should say so.

cmd/herdr-auto-title/main.go
```diff
@@ -36,18 +52,85 @@ func run() error {
...
+			"the session could not be claimed, so a restart will not end this instance",
+			"error",
+			err,
+		)
+	}
+	defer claim.Release()
+
+	// The displaced instance looks at the claim before every poll, and leaves
+	// on its own; the manual-name locks are loaded only once it has.
+	if old := claim.Displaced(); old != 0 {
```
kryptamine: `Displaced()` and `AwaitDisplaced()` are public only for this block, and `old != 0` is checked both here and inside `AwaitDisplaced`. `Take` could wait for the predecessor itself, taking a `ctx` and a timeout and returning the pid if it didn't leave. `main` keeps its log lines and `Claim` loses two public methods.

internal/instance/instance.go
```diff
@@ -0,0 +1,205 @@
...
+
+		return false
+	}
+
+	return rec.PID != c.pid
+}
+
+// Ready marks this instance as having read the session, in the marker rather
+// than the claim, so a claim taken meanwhile is never written over.
+func (c *Claim) Ready() {
```
kryptamine, line 97: `poll` calls this after every successful snapshot, so it reads the marker file on every poll just to learn it has already written it. A `ready bool` on `Claim` does the same job.

internal/herdr/client.go
```diff
@@ -155,3 +160,20 @@ func RenameTab(ctx context.Context, c Client, tabID, label string) error {
 func RenamePane(ctx context.Context, c Client, paneID, label string) error {
 	return c.Call(ctx, MethodPaneRename, PaneRenameParams{PaneID: paneID, Label: label}, nil)
 }
+
+// ShowNotification puts a notice in front of the user and reports whether it
+// was shown, with Herdr's reason when it was not: nothing is shown without a
+// client attached, and two within a second is one too many.
+func ShowNotification(ctx context.Context, c Client, title, body string) (bool, string, error) {
```
kryptamine: This returns `(bool, string, error)`, but `shown` and `reason` only feed a debug log in `main`. Returning an `error` and logging the reason inside, or returning the result struct, is simpler.

internal/instance/instance.go
recih: You are right, and the race is real: `superseded` is checked before `Taken`, so the displaced instance leaves through the socket check and its `Release` removes the claim the successor has just written.

`Release` is gone, and with it `exists`, `names` and the repair write in `Taken`. `Taken` is now a read and a comparison. A claim naming a dead pid displaces nobody, which is how a crashed instance was already handled, so the only case lost is a pid reused after a clean exit — a wait that times out and a warning.

`TestAnInstanceLeavingKeepsItsSuccessorsClaim` covers it end to end: a real instance is started, displaced, and once it has exited the claim still names this process.

internal/instance/restart.go
recih: Done. `pending(path, pid, old) error` returns the first unmet condition or nil. The wait is `pending(...) == nil`, the deadline reports `gave up after %s: %w` with the same error, and `stalled` is gone.

The second `holder()` call went too: `ready(path)` became `marked(path, pid)`, a plain marker read, since `pending` has already established who holds the claim.

internal/instance/instance.go
recih: Gone with `Release`. `Taken` is now:

```go
func (c *Claim) Taken() bool {
	rec, ok := read(c.path)

	return ok && rec.PID != c.pid
}
```

A missing claim is nobody's and ends no run; nothing outside `Take` writes one. The paragraph in poll-loop.md that documented the pairing went with it.

internal/instance/instance.go
recih: Done: `waitUntil(ctx, timeout, done func() bool) bool` backs both. `Take` passes `func() bool { return !alive(displaced) }`, and the restart reads `exited` without blocking inside its condition:

```go
done := waitUntil(ctx, timeout, func() bool {
	select {
	case state = <-exited:
		return true
	default:
		return pending(path, proc.Pid, old) == nil
	}
})
```

The caller tells a cancelled context from a deadline through `ctx.Err()`.

internal/instance/restart.go
recih: Agreed. `restartTimeout` is gone and `Restart` takes the budget:

```go
pid, err := instance.Restart(ctx, instance.File(app.StateDir(), client.Path()), exe, app.LeaveTimeout+app.PollTimeout)
```

The private `restart(..., timeout)` is gone with it — the tests call `Restart` directly.

internal/app/app.go
recih: `poll` has one exit now:

```go
if reason := a.successor(client); reason != "" {
	a.log.Info(reason + ", leaving")
	return false
}
```

`successor` asks the claim first and the socket second. I kept the socket check rather than deleting it: it is the only answer left when there is no configuration directory to claim in, so an instance that could not claim still leaves when the server changes. That is what the comment and poll-loop.md now say. Asking the claim first also removes the ordering that opened the `Release` race, though with `Release` gone there is nothing left to race.

cmd/herdr-auto-title/main.go
recih: Done. `Take(ctx, path, leave) (*Claim, int, error)` waits for the predecessor itself and returns the pid that stayed, or 0. `Displaced` and `AwaitDisplaced` are gone and `old != 0` is checked once, inside `Take`.

One log line did not survive: `main` used to say "waiting for the instance this one replaces to leave" before the wait, which it can no longer know about. The warning when one stays is still there. Happy to pass a callback or return the displaced pid too if you would rather keep it.

internal/instance/instance.go
recih: Done — `Claim` carries a `ready bool`, set from the write, and `Ready` returns at once on every later poll. Nothing reads the marker back in the running instance now; only the restart does, through `marked(path, pid)`.

internal/herdr/client.go
recih: `notificationResult` is now exported and returned whole:

```go
func ShowNotification(ctx context.Context, c Client, title, body string) (NotificationResult, error)
```

`main` reads `notice.Shown` and `notice.Reason` for the debug line. Logging inside would have meant giving the package a logger, so I took the struct.

recih:
Thanks — all nine are addressed, and the branch is rebased onto `main` at v0.7.0 (the XDG change touched `ConfigDir`, which is now `app.StateDir` in `paths.go`: always the platform directory, never `ownPath`'s first match, so every instance agrees on where the claim is).

The one behavioural change is the first comment: an instance no longer removes anything on its way out. `Release`, `exists`, `names` and the repair write in `Taken` are all gone, `Taken` is a read and a comparison, and a new test drives the race you described with real processes.

The rest are the simplifications you asked for: one `pending()` behind both the wait and its report, one `waitUntil()` behind both waits, the restart budget passed in from `main`, `Take` waiting for its predecessor itself, a `ready bool` on `Claim`, and `ShowNotification` returning `NotificationResult`.

`make check` is green, and I ran it against a live Herdr 0.9.1 on Windows: the action returns at once, a second restart displaces the first instance, and once that one has exited the claim and the marker both name the survivor.

It is amended into the one commit rather than added on top, so the diff you reviewed has been rewritten — sorry for that. Two things I would like your call on: the log line noted under `main.go`, and whether the socket check earns its keep now that the claim covers a new server.

kryptamine:
I tried it on macOS with Herdr 0.9.0, from the CLI and from a key, and it does what the issue asks. I pushed a few commits on top of yours. One is a code fix: two invocations within half a second made the first one report a failure, although its instance had exited only because the second one's instance had claimed the session. The rest is documentation: a note that the notification needs Herdr's toasts on, the `Restarting` section moved below `Configuration`, and a few lines in `development.md` that no longer matched the code.

kryptamine (approved):

kryptamine:
LGTM, merging this now. Thanks!

PR #93, opened 2026-09-17
feat(app): honour XDG_CONFIG_HOME and ~/.config

### LINKED ISSUE #81: Respect XDG_CONFIG_HOME and use ~/.config on macOS too
### The problem

On macOS the config file lives at `~/Library/Application Support/herdr-auto-title/config.env`. On Linux it lives at `~/.config/herdr-auto-title/config.env`.

I sync my dotfiles across machines, as a lot of people do. `~/.config` is in that sync; `~/Library/Application Support` is not, and it is a macOS-only path that does not belong in a shared dotfiles repo. So the same setting has to be kept in two places, one of which I cannot version.

### What it should do instead

Nothing changes in the tab bar. This is a setting-location change.

Look for `config.env` in this order:

1. `$XDG_CONFIG_HOME/herdr-auto-title/config.env` when `XDG_CONFIG_HOME` is set.
2. `~/.config/herdr-auto-title/config.env`.
3. `~/Library/Application Support/herdr-auto-title/config.env` on macOS, so existing installs keep working.

The same order would apply to `HERDR_AUTO_TITLE_MANUAL_FILE`, which defaults to `manual-names.json` next to `config.env`.

On my machine:

```
$ echo $XDG_CONFIG_HOME
/Users/ian/.config
```

`XDG_CONFIG_HOME` is the cross-platform answer and it already covers the Linux case, so honouring it first means one rule instead of a per-OS table.

The `XDG_CONFIG_HOME` environment variable, falling back to `$HOME/.config`, then the current macOS path.

### What you tried

Symlinking `~/Library/Application Support/herdr-auto-title` at the copy in my dotfiles. It works, but it is one more thing to set up on every new Mac, and it is invisible to anyone reading the dotfiles repo.

Thanks!

kryptamine:
Closes #81.

## The problem

The configuration directory was whatever the platform offers alone, which on
macOS is `~/Library/Application Support`: a macOS-only path that cannot go in a
dotfiles repository. A user with a Mac and a Linux box kept the same setting in
two files and could version only one copy of it.

## What changes

Each of Auto Title's files is now looked for in three directories, and the
first that **already holds it** wins:

| Order | Directory |
|-------|-----------|
| 1 | `$XDG_CONFIG_HOME`, when it is set to an absolute path |
| 2 | `~/.config`, which is `%USERPROFILE%\.config` on Windows |
| 3 | `os.UserConfigDir()`: `~/Library/Application Support` on macOS, `%APPDATA%` on Windows |

- **Nothing is moved.** An existing install keeps reading its file where it is;
  putting one in `~/.config` is how a user opts in, and deleting it is how they
  go back.
- **Each file resolves on its own**, so moving only `config.env` into a dotfiles
  repository leaves `manual-names.json` where it was — locks are machine state,
  not configuration.
- **A new file goes to the last directory**, `os.UserConfigDir()`, not the
  first, so a fresh install does not start a `~/.config` for its locks. On
  Linux that directory is `$XDG_CONFIG_HOME` or `~/.config` anyway.
- A relative `XDG_CONFIG_HOME` is ignored, as the specification says: honoured,
  it would resolve against whatever directory the Herdr server was started in.
- `state.DefaultManualPath` goes with its last reader; `internal/app` now owns
  where the plugin's files live, in `internal/app/paths.go`.

## Verified on macOS, against a live session

| Case | Result |
|------|--------|
| Existing `config.env` in `Application Support` | read as before |
| `config.env` in `~/.config` | wins over `Application Support` |
| Locks with `config.env` moved to `~/.config` | stay in `Application Support` |
| `XDG_CONFIG_HOME` set | read first |
| `XDG_CONFIG_HOME` relative | ignored, falls back to `~/.config` |

`make check` is green, and seven tests in `internal/app/paths_test.go` cover the
ordering and where a new file lands; the ones that need a platform directory
apart from `~/.config` skip on Linux, which has none.

Docs: `docs/architecture/configuration.md` records why the order is what it is
and why the third entry is a fallback rather than a legacy to migrate; the
README and `config.env.example` say where to put the file.


kryptamine:
@recih @6temes guys, please take a look when you have a chance. Thanks!

recih:
**`~/.config` exists on Windows too**

The PR skips `~/.config` on Windows because "Windows has no equivalent of" it. It does: it's `%USERPROFILE%\.config`, and a lot of cross-platform tools already keep their configuration there on Windows. On my machine that includes git, mise, scoop, wezterm and opencode. A dotfiles repository that serves Windows too syncs that directory. That is the same reason #81 gives for wanting `~/.config` on macOS.

`os.UserHomeDir()` returns `%USERPROFILE%` on Windows, so I think the only code change is dropping the platform check:

```go
if home, err := os.UserHomeDir(); err == nil {
	dirs = append(dirs, filepath.Join(home, ".config"))
}
```

On Windows the order would then be `XDG_CONFIG_HOME`, then `%USERPROFILE%\.config`, then `%APPDATA%`. Existing installs are unaffected, because a file already in `%APPDATA%` still wins when `.config` holds none.

It would also mean:
- `dotConfig` in `paths_test.go` no longer needs to skip on Windows. `setHome` already sets `USERPROFILE`, so the tests stay isolated. The ordering tests would then run on the Windows job as well as on macOS.
- The README, `config.env.example` and `configuration.md` would drop "which Windows has no equivalent of", and the paragraph starting "Windows keeps `%APPDATA%` and gains nothing but the variable". Herdr keeping its own files under `%APPDATA%\herdr` is Herdr's choice for its own directory; it doesn't need to decide where a user's dotfiles live.

kryptamine:
Thanks! I didn't know that `~/.config` is also used on Windows. I'll fix this.

internal/app/paths.go
```diff
@@ -0,0 +1,68 @@
...
+		dirs = append(dirs, xdg)
+	}
+
+	// ~/.config is the Unix convention macOS shares and Windows has no
+	// equivalent of, so only Windows is left with its platform directory.
+	if runtime.GOOS != "windows" {
+		if home, err := os.UserHomeDir(); err == nil {
+			dirs = append(dirs, filepath.Join(home, ".config"))
+		}
+	}
```
recih: `%USERPROFILE%\.config` is the Windows counterpart (see my comment above), so this could apply everywhere:

```suggestion
	// ~/.config is where cross-platform tools keep configuration, Windows
	// included, as %USERPROFILE%\.config.
	if home, err := os.UserHomeDir(); err == nil {
		dirs = append(dirs, filepath.Join(home, ".config"))
	}
```

The `runtime` import then goes too.

kryptamine: fixed

kryptamine:
@recih pushed the changes, ready for another look when you get a chance. Could you please also build the branch and test it on windows? You're on windows, right?

6temes (commented):
Read it at `e25b53e`, `make check` green here, lint 0 issues. The ordering is
right and the reasoning in `configuration.md` for not migrating anything is the
part I would have argued for. One thing I think contradicts that reasoning.

## A fresh install puts its locks in the dotfiles directory

`configuration.md` says it plainly, and I agree with it:

> locks are machine state, not configuration, and a user syncing dotfiles is
> not asking for a Mac's tab locks on a Linux box

But `ownPath` falls back to the **first** directory when none holds the file,
and `~/.config` is added unconditionally — no check that it exists. So on a Mac
with no `XDG_CONFIG_HOME` and no `~/.config`, a new `manual-names.json` is
created at `~/.config/herdr-auto-title/manual-names.json`, and `~/.config` is
created for it. `~/Library/Application Support` stays empty.

Measured on this branch, in an isolated home with no `~/.config`:

```
ManualPath = <home>/.config/herdr-auto-title/manual-names.json
~/.config copy exists: true
platform copy exists:  false
```

So the "each file is looked for on its own" argument holds for an install that
already has a lock file, and inverts for one that does not: the first thing a
fresh Mac install does is put machine state in the directory the document
describes as the one a dotfiles repository syncs. A user who later adopts
`~/.config` for dotfiles finds their tab locks already committed there.

Two ways out, both small:

1. Resolve the **write** preference differently from the read order: look in all
   three, but when none holds the file, prefer `os.UserConfigDir()`. Reading
   stays one rule; only the fallback changes.
2. Keep one list, but only offer `~/.config` as the fallback when it already
   exists as a directory. A user who keeps dotfiles has one; a user who does not
   should not have one created to hold locks.

I prefer 1, because it keeps the fallback independent of what happens to be on
disk, and because it makes the `config.env` case and the `manual-names.json`
case differ for the reason the document already gives.

Either way the six tests in `paths_test.go` stay green — they all cover reading
a file that exists, and none pins where a new one lands, which is why this is
invisible to the suite.

## Smaller things

- The PR body's table still reads "`~/.config`, which Windows has no equivalent
  of", which `e25b53e` reverses. `README.md` and `config.env.example` both have
  it right; only the body is stale.
- `configuration.md` going from nine variables to ten is a fix to a count that
  was already stale on `main`, not a change this makes — worth knowing, since
  #89 makes it eleven.

## Overlap with #89

This and #89 both touch `internal/app/config.go`, `config_test.go`,
`README.md`, `config.env.example` and `docs/architecture/configuration.md`.
Whichever lands second needs the rebase and I am happy for that to be mine —
say which order you want and I will rebase #89 onto this.

recih:
@kryptamine I ran the branch on Windows 11 to check the Windows side. `make check` is green there.

One thing worth reporting: **three of the six path tests skip on Windows**, so the Windows job covers less than it looks.

```
--- SKIP: TestTheHomeConfigDirectoryIsLookedIn
--- SKIP: TestTheHomeConfigDirectoryBeatsThePlatformOne
--- SKIP: TestEachFileIsLookedForOnItsOwn
        Windows keeps configuration in %AppData%, not in ~/.config
```

Dropping the platform check in `configDirs` and the matching `t.Skip` in `dotConfig` makes all three run and pass on Windows with no other change — `setHome` already sets `USERPROFILE`, so the tests stay isolated. The ordering would then be covered by the Windows job as well as the macOS one, and an existing `%APPDATA%` install still wins, exactly as the PR intends.

kryptamine:
@6temes Thanks, good catch!

I'll go with option 1: read from all three dirs as before, but create a missing file in `os.UserConfigDir()`. I'll add a test for that and fix the stale line in the PR body.

Let's merge this one first, then you rebase #89 on top. Thanks for offering!

kryptamine:
@recih Thanks for running it on Windows! That's already in e25b53e: `configDirs` no longer checks the platform and the skip in `dotConfig` is gone, so all three run on Windows now. Looks like you tested the commit before it.

I'm fixing @6temes's review comments right now, so once that's pushed, could you give it another run on Windows?

kryptamine:
Review comments are addressed in 90a8e2d. Ready for another look!

recih:
@kryptamine You're right, I had fetched the branch before e25b53e, sorry for the noise. I ran 90a8e2d on Windows 11 this time: `make check` is green, and all seven path tests run and pass with none skipped, `TestANewFileGoesToThePlatformDirectory` included. An existing `%APPDATA%` install is still read where it is, and a missing file is created under `%APPDATA%`.

Heads-up: #92 also touches `configPath()` in `internal/app/config.go`, so it will conflict with this one. Go ahead and merge this first; I'll rebase #92 on top afterwards.

recih (approved):
LGTM

kryptamine:
@6temes I'm merging this. If you have any further comments, let me know and I'll address them in a separate PR.

PR #96, opened 2026-09-20
fix(resolver): treat "Windows PowerShell" as a generic title

2451965602:
Summary`n- treat Windows PowerShell as a generic terminal title for every pane`n- cover the multi-word Windows shell spelling in the generic-title table`n`n## Testing`n- make check


kryptamine:
Thanks for the contribution!
`genericValues` (`internal/resolver/generic.go:11`) is the table of values that name a program or a place instead of the work, and any source that finds one declines. It already holds `"node"`, `"shell"`, `"claude"` and the multi-word `"claude code"`. `TestMeaningful` covers both spellings this bug is about:

```go
{"the older windows shell", "PowerShell", "", false},   // generic_test.go:31
{"multi-word program name", "Claude Code", "", false},  // generic_test.go:33
```

So `PowerShell` is refused today for every pane. `Windows PowerShell` gets through only because the multi-word spelling was never added. The fix is one row:

```go
 var genericValues = map[string]struct{}{
 	// Terminals and runtimes.
-	"shell":    {},
+	"shell":              {},
+	"windows powershell": {},
 	"terminal": {},
```



@recih could you please take a look at this one as well?

recih (commented):
Thanks for the contribution. The code change is right: `"windows powershell"` belongs in `genericValues`, next to the single-word `PowerShell` that is already refused. There are a few things around it I'd change before this merges:

1. **The commit message no longer matches the change.** The fix now applies to every pane whose title is `Windows PowerShell`, not only to OpenCode panes. This repository merges by rebase, so the subject goes into the changelog unchanged. Something like this would fit:

   ```
   fix(resolver): treat "Windows PowerShell" as a generic title

   Windows PowerShell titles its console "Windows PowerShell", and only
   the single-word spelling was refused, so a pane whose program set no
   title of its own was named after the shell.
   ```

   The PR title should say the same.

2. **Please drop the paragraph added to `docs/architecture/title-resolution.md`.** It's about not reading OpenCode's private store, which this PR doesn't change. It also sits in the section on Claude Code transcripts. And "until Herdr reports a title for that session" isn't accurate, because Herdr doesn't report session titles. If OpenCode session titles are worth tracking, that belongs in an issue of its own.

3. **The new app test describes a generic behaviour as an OpenCode one.** A pane falling back to its agent name when the terminal title isn't usable already worked before this change. What this PR fixes is covered by the new `TestMeaningful` row. I'd remove `TestAnOpenCodePaneWithNoTitleFallsBackToTheAgentName`. If you'd like to keep an end-to-end case, rename it so it says what it protects, e.g. that a `Windows PowerShell` title never becomes a tab name.

With those changes this looks good to me.

2451965602:
Addressed in 9f23fa5: the fix now only adds `windows powershell` to `genericValues` with a `TestMeaningful` case. I removed the unrelated OpenCode documentation and redundant app test, and updated the commit and PR title. `make check` passes.

recih (approved):

kryptamine (approved):

